PR #11, opened 2023-01-23
wallet: Implement ADR14

matevz:
Requires https://github.com/oasisprotocol/oasis-sdk/pull/1248

- Adds support for signing runtime transactions with Ledger
- Bumps oasis-sdk from 0.3.0 to 0.4.0
- Adds Migration primitives for wallet config

### DIFF
--- a/cmd/common/transaction.go
+++ b/cmd/common/transaction.go
@@ -147,6 +147,7 @@ func SignParaTimeTransaction(
 	wallet wallet.Account,
 	conn connection.Connection,
 	tx *types.Transaction,
+	txDetails *signature.TxDetails,
 ) (*types.UnverifiedTransaction, interface{}, error) {
 	// Default to passed values and do online estimation when possible.
 	nonce := txNonce
@@ -240,12 +241,16 @@ func SignParaTimeTransaction(
 	PrintTransactionBeforeSigning(npa, tx)
 
 	// Sign the transaction.
-	sigCtx := signature.DeriveChainContext(npa.ParaTime.Namespace(), npa.Network.ChainContext)
 	ts := tx.PrepareForSigning()
+	sigCtx := &signature.RichContext{
+		RuntimeID:    npa.ParaTime.Namespace(),
+		ChainContext: npa.Network.ChainContext,
+		Base:         types.SignatureContextBase,
+		TxDetails:    txDetails,
+	}
 	if err := ts.AppendSign(sigCtx, wallet.Signer()); err != nil {
 		return nil, nil, fmt.Errorf("failed to sign transaction: %w", err)
 	}
-
 	return ts.UnverifiedTransaction(), meta, nil
 }
 
--- a/cmd/common/transaction.go
+++ b/cmd/common/transaction.go
@@ -268,12 +268,6 @@ func PrintTransaction(npa *NPASelection, tx interface{}) {
 		coreSignature.UnsafeResetChainContext()
 		coreSignature.SetChainContext(npa.Network.ChainContext)
 		rtx.PrettyPrint(ctx, "", os.Stdout)
-	case *types.Transaction, *types.UnverifiedTransaction:
-		// TODO: Add pretty variant for paratime transactions.
-		formatted, err := json.MarshalIndent(tx, "", "  ")
-		cobra.CheckErr(err)
-		fmt.Println(string(formatted))
-		isParaTimeTx = true
 	default:
 		fmt.Printf("[unsupported transaction type: %T]\n", tx)
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -161,6 +161,16 @@ func (cfg *Config) Save() error {
 	return cfg.viper.WriteConfig()
 }
 
+// Migrate migrates the given wallet config entry to the latest version and returns true, if any changes were needed.
+func (cfg *Config) Migrate() (bool, error) {
+	changes, err := cfg.Wallet.Migrate()
+	if err != nil {
+		return false, fmt.Errorf("failed to migrate wallet configuration: %w", err)
+	}
+
+	return changes, nil
+}
+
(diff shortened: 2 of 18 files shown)


cmd/common/transaction.go
```diff
@@ -240,9 +241,18 @@ func SignParaTimeTransaction(
 	PrintTransactionBeforeSigning(npa, tx)
 
 	// Sign the transaction.
-	sigCtx := signature.DeriveChainContext(npa.ParaTime.Namespace(), npa.Network.ChainContext)
+	// If hardware wallet used follow signing procedure as defined in ADR 14.
 	ts := tx.PrepareForSigning()
-	if err := ts.AppendSign(sigCtx, wallet.Signer()); err != nil {
+	var err error
+	if wallet.UnsafeExport() == "" {
```
kostko, line 247: This doesn't seem like the best API as the caller now needs to handle an internal detail of the underlying Ledger protocol (e.g. CBOR serialization of the metadata field).

Could we instead rename `HwSignRtMeta` to something like `ExtendedContext` (given that it lives in the context) and then make the signers accept a new context interface for the ctx, e.g.:
```golang
type SignContext interface {
    // Extended returns the extended signing context if available. Otherwise it returns nil.
    Extended() *ExtendedContext

    // Derive derives the raw signing context.
    Derive() []byte
}
```

This interface would be implemented for both the existing `Context` type (where `Extended()` could just return `nil`) and the new `ExtendedContext` type (where `Extended()` would return itself and `Derive` would use the `DeriveChainContext` method).

Then you could propagate this new interface all the way to the `Signer` and the Ledger implementation can then try calling `.Extended()` to get the extended context and then handle serialization there. Other signers (e.g. file) can use `Derive()` to get the underlying context.

wallet/ledger/device.go
```diff
@@ -2,11 +2,18 @@ package ledger
 
 import (
 	"fmt"
+	"math/big"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec"
```
kostko, line 8: Please switch to `github.com/btcsuite/btcd/btcec/v2` (see the latest SDK that also switched).

wallet/ledger/ledger.go
```diff
@@ -87,14 +86,33 @@ func (af *ledgerAccountFactory) SupportedImportKinds() []wallet.ImportKind {
...
+	}
+	return false
 }
 
 func (af *ledgerAccountFactory) unmarshalConfig(raw map[string]interface{}) (*accountConfig, error) {
 	if raw == nil {
 		return nil, fmt.Errorf("missing configuration")
 	}
 
+	// CONFIG MIGRATION 1 (add legacy derivation support): Set default derivation to ADR 8, if not set.
```
kostko, line 106: Will this ever be saved back? It would be good if the first time migration is performed the underlying config is actually updated.

kostko (commented):
Nice, looks mostly good, just some nits and it should be ready to merge!

wallet/ledger/ledger.go
```diff
@@ -87,14 +86,50 @@ func (af *ledgerAccountFactory) SupportedImportKinds() []wallet.ImportKind {
...
+	return changed
 }
 
 func (af *ledgerAccountFactory) unmarshalConfig(raw map[string]interface{}) (*accountConfig, error) {
 	if raw == nil {
 		return nil, fmt.Errorf("missing configuration")
 	}
 
+	if changed := af.migrate(raw); changed {
+		config.Global().Save()
```
kostko, line 130: I worry this could have unintended side effects as there could be other configuration changes pending that should not be saved. Could we make a separate `Migrate(name, oldCfg) newCfg` step on the factory which would perform any migrations on the accounts? And then we could have a migration step in the `initConfig` phase (see `cmd/root.go`).

matevz: I did something similar now. Can you check it out?

config/config.go
```diff
@@ -8,9 +8,8 @@ import (
 	"strings"
 
 	"github.com/adrg/xdg"
-	"github.com/spf13/viper"
-
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+	"github.com/spf13/viper"
```
kostko, line 12: This grouping looks wrong.

config/wallet.go
```diff
@@ -18,6 +17,20 @@ type Wallet struct {
 	All map[string]*Account `mapstructure:",remain"`
 }
 
+func (w *Wallet) Migrate() (bool, error) {
```
kostko, line 20: Add documentation comment.

config/wallet.go
```diff
@@ -94,6 +107,11 @@ func (w *Wallet) Load(name string, passphrase string) (wallet.Account, error) {
 		return nil, err
 	}
 
+	// Run any account migration and save the config.
```
kostko, line 110: Why would this be needed here if global migrations ran during config init?

wallet/ledger/ledger.go
```diff
@@ -87,7 +85,39 @@ func (af *ledgerAccountFactory) SupportedImportKinds() []wallet.ImportKind {
...
+	}
+
+	// CONFIG MIGRATION 2 (convert derivation -> algorithm).
+	if val, ok := rawCfg["derivation"]; ok {
+		rawCfg["algorithm"] = fmt.Sprintf("ed25519-%s", val)
+		delete(rawCfg, "derivation")
+
+		changed = true
+	}
+	fmt.Println(rawCfg)
```
kostko, line 119: Remove?

config/wallet.go
kostko: Or at least do not save in this case.

matevz: Just a leftover, sorry.

kostko (approved):

PR #14, opened 2023-01-26
Add basic versioning and release workflow

kostko:


version/version.go
```diff
@@ -0,0 +1,33 @@
...
+	// Software represents the Oasis CLI version and should be set by the linker.
+	Software = "0.0.0-unset"
+
+	// Toolchain is the version of the Go compiler/standard library.
+	Toolchain = runtime.Version()
+)
+
+// GetOasisSDKVersion returns the version of the Oasis SDK dependency or unknown if it can't be
+// obtained.
+func GetOasisSDKVersion() string {
```
matevz, line 21: We import some stuff from oasis-core package directly too. Should we also fetch the version of that?

matevz (approved):

PR #47, opened 2023-03-28
feat(cmd): Add support for registry operations

kostko:


cmd/registry.go
```diff
@@ -0,0 +1,223 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	"github.com/oasisprotocol/oasis-core/go/common/entity"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+
```
matevz, line 14: I usually group core and sdk together.

```suggestion
```

matevz (changes requested):
Few questions:
1. Where is nodeRegister operation?
2. Should we move `oasis inspect registry` to something like `oasis registry show` instead? Also`oasis registry dump` would come in handy.

kostko:
Node registration happens automatically by oasis-node so manual registration should never be needed.

Good point about the registry inspect functions. By dump you probably mean something like the current `inspect registry nodes`?

cmd/registry.go
kostko: Fixed.

kostko:
Moved the `inspect registry` part as well.

cmd/registry.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+
+			// Load the account and ensure it corresponds to the entity.
+			acc := common.LoadAccount(cfg, npa.AccountName)
+			signer := acc.ConsensusSigner()
+			if signer == nil {
+				cobra.CheckErr("account does not support signing consensus transactions")
+			}
+			if !signer.Public().Equal(descriptor.ID) {
+				cobra.CheckErr(fmt.Errorf("entity ID (%s) does not correspond to selected account (%s)",
```
matevz, line 70:
```suggestion
				cobra.CheckErr(fmt.Errorf("entity ID '%s' does not correspond to selected account '%s' (%s)",
```

cmd/registry.go
```diff
@@ -0,0 +1,419 @@
...
+					descriptor.ID, signer.Public()))
```
matevz, line 71:
```suggestion
					descriptor.ID, npa.AccountName, signer.Public()))
```

cmd/registry.go
```diff
@@ -0,0 +1,419 @@
...
+			var descriptor entity.Entity
+			if err = json.Unmarshal(rawDescriptor, &descriptor); err != nil {
+				cobra.CheckErr(fmt.Errorf("malformed entity descriptor: %w", err))
+			}
+
+			// Load the account and ensure it corresponds to the entity.
+			acc := common.LoadAccount(cfg, npa.AccountName)
+			signer := acc.ConsensusSigner()
+			if signer == nil {
+				cobra.CheckErr("account does not support signing consensus transactions")
```
matevz, line 67:
```suggestion
				cobra.CheckErr(fmt.Errorf("account '%s' does not support signing consensus transactions", npa.AccountName))
```

cmd/registry.go
```diff
@@ -0,0 +1,419 @@
...
+				var err error
+				conn, err = connection.Connect(ctx, npa.Network)
+				cobra.CheckErr(err)
+			}
+
+			// Prepare transaction.
+			tx := registry.NewDeregisterEntityTx(0, nil)
+
+			acc := common.LoadAccount(cfg, npa.AccountName)
+			sigTx, err := common.SignConsensusTransaction(ctx, npa, acc, conn, tx)
```
matevz, line 116: It would be nice if PrettyPrinter omitted the

```
Body:
  {}
```

chunk if Body is empty. But this needs to be fixed in SDK.

cmd/registry.go
```diff
@@ -0,0 +1,419 @@
...
+	registryEntityDeregisterCmd.Flags().AddFlagSet(common.SelectorFlags)
+	registryEntityDeregisterCmd.Flags().AddFlagSet(common.TransactionFlags)
+
+	registryNodeUnfreezeCmd.Flags().AddFlagSet(common.SelectorFlags)
+	registryNodeUnfreezeCmd.Flags().AddFlagSet(common.TransactionFlags)
+
+	registryRuntimeRegisterCmd.Flags().AddFlagSet(common.SelectorFlags)
+	registryRuntimeRegisterCmd.Flags().AddFlagSet(common.TransactionFlags)
+
+	registryShowCmd.Flags().AddFlagSet(common.SelectorFlags)
```
matevz, line 411: `oasis registry show`  only requires `--network` flag afaik.

matevz (approved):
Good job!

matevz:
Ah, crap, automerge was enabled. Can you make a short followup?

kostko:
Yeah will do.

kostko:
@matevz should be addressed in #49.

PR #56, opened 2023-04-12
ci: undisable cgo on release build

pro-wh:
commit message
> without cgo, we can't use ledger. the library does not link to libusb and always answers that no devices are connected.
https://github.com/Zondax/hid/blob/v0.9.1/hid_disabled.go#L7
this CLI is intended to work with ledger, so we're enabling this.

we're not aware of anything that specifically needed a non-cgo build. but if anyone depended on the program not being with cgo, you're welcome to comment.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -6,8 +6,6 @@ before:
 
 builds:
   - binary: oasis
-    env:
-      - CGO_ENABLED=0
     flags:
       - -trimpath
     ldflags:


.goreleaser.yaml
```diff
@@ -6,8 +6,6 @@ before:
 
 builds:
   - binary: oasis
-    env:
-      - CGO_ENABLED=0
```
kostko, line 10: I believe the only reason for having this was that this config was copied from the [oasis-web3-gateway](https://github.com/oasisprotocol/oasis-web3-gateway) that had this set.

kostko (approved):

.goreleaser.yaml
pro-wh: as long as folks aren't using this in a no-libc container or something :D

PR #67, opened 2023-05-09
feat(cmd): Add support for outputting unsigned transactions

kostko:


cmd/common/transaction.go
```diff
@@ -338,10 +354,33 @@ func PrintTransactionBeforeSigning(npa *NPASelection, tx interface{}) {
 
 // PrintSignedTransaction prints a signed transaction.
 func PrintSignedTransaction(sigTx interface{}) {
-	// TODO: Add some options for controlling output.
```
matevz, line 341: `Print*Transaction` method names seem outdated. So `PrintSignedTransaction()` is now used to print unsigned transactions as well? I would refactor the code a bit so `BroadcastTransaction` is not called in offline and txUnsigned mode anymore (do we want to broadcast unsigned tx in any scenario?)

matevz (changes requested):

cmd/common/transaction.go
kostko: Renamed `PrintSignedTransaction` to `ExportTransaction` as that seems more appropriate now.

Also there are now separate `BroadcastTransaction` and `BroadcastOrExportTransaction` and usage has been updated.

matevz (changes requested):
Thanks for the code cleanup. I tested this and usage seems awkward. Example:

```
oasis account transfer 1.0 test:bob
```

will sign and broadcast the transaction.

One would expect that using the new `-o` switch would dump the tx to a file instead of broadcasting it:

```
oasis account transfer 1.0 test:bob -o transfer-bob.json
```

But it still broadcasts the transaction (and `-o` switch does nothing).

So what you do is put an `--offline` flag, but then you get

```
$ oasis account transfer 1.0 test:bob --offline -o transfer-bob.json
Error: nonce and/or gas limit must be specified in offline mode
```

I suggest the following: `-o` switch should mean "output tx to file instead of broadcasting it". But this should be separate from the `--offline` switch which should mean "don't use network to determine transaction parameters".

kostko:
Fixed, please take another look.

matevz (commented):
I was trying to generate an unsigned transactio and then sign and submit it. Are two authorized signers expected here?

```
$ oasis account transfer 2.5 test:bob -o test.tx  --unsigned
Unlock your account.
? Passphrase: 
oa@matevz-oa:~/cli$ oasis tx submit test.tx 
Unlock your account.
? Passphrase: 
You are about to sign the following transaction:
Format: plain
Method: accounts.Transfer
Body:
  To: test:bob (oasis1qrydpazemvuwtnp3efm7vmfvg3tde044qg6cxwzx)
  Amount: 2.5 TEST
Authorized signer(s):
  1. MJ2XCjkj132C9YWpDUSQFjkCTI8bSw8bi0w9EwwE1Bg= (ed25519)
     Nonce: 3
  2. MJ2XCjkj132C9YWpDUSQFjkCTI8bSw8bi0w9EwwE1Bg= (ed25519)
     Nonce: 3
Fee:
  Amount: 0.001152295 TEST
  Gas limit: 230459
  (gas price: 0.000000005 TEST per gas unit)

Network:  testnet
ParaTime: cipher
Account:  test
? Sign this transaction? (y/N)
```

matevz (approved):
Good job!

Issue #73, opened 2023-05-11
Merge `inspect` and `registry` commands with `network`, `paratime`, `account` commands


matevz:
Current `inspect` and `registry` platform-specific commands seem out of place compared to `account`, `paratime` and `network` commadns which are user-centric. As a rule-of-thumb, entity and validator operations go under `account`, consensus operations go under `network` and runtime operations go under `paratime`. I suggest we merge those accordingly:
- `oasis registry runtime-register` -> `oasis paratime register`
- `oasis registry entity-register`, `oasis registry entity-deregister`, `oasis registry node-unfreeze` could go under `oasis account entity register`, `oasis account entity deregister`, `oasis account node-unfreeze`
- `oasis registry show` -> `oasis network show`
- `oasis inspect native-token` -> `oasis network info` (in the future, this could also be added to `oasis paratime` for querying ParaTime-native tokens, but currently this is not supported yet anyway)
- `oasis inspect node-status` -> `oasis network status`
- `oasis inspect runtime-stats` -> `oasis paratime stats`
- `oasis inspect governance-proposal` -> `oasis network governance-proposal`
- `oasis inspect block` -> `oasis paratime show`

PR #75, opened 2023-05-16
Refactor network, paratime and account commands

matevz:
Implements #73 .

Also:
- splits `account.go` into `account/` folder containing file for each subcommand
- splits `network.go` into `network/` folder containing file for each subcommand
- splits `paratime.go` into `paratime/` folder containing file for each subcommand
- adds gas costs and debonding interval to `network info`
- adds selectors for network-only and network-account-only


kostko (commented):
Initial comments, haven't reviewed everything yet.

cmd/account/transfer.go
```diff
@@ -0,0 +1,101 @@
...
+		common.CheckForceErr(common.CheckAddressNotReserved(cfg, toAddr.String()))
+
+		acc := common.LoadAccount(cfg, npa.AccountName)
+
+		var sigTx, meta interface{}
+		switch npa.ParaTime {
+		case nil:
+			if txCfg.Encrypted {
+				cobra.CheckErr("--encrypted not supported for consensus transactions")
+			}
```
kostko, line 56: You could make this general and put the check into `SignConsensusTransaction` so it is always performed?

cmd/account/entity.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func init() {
+	entityRegisterCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	entityRegisterCmd.Flags().AddFlagSet(common.TxFlags)
+
+	entityDeregisterCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	entityDeregisterCmd.Flags().AddFlagSet(common.TxFlags)
+
+	entity := entity.Entity{
```
kostko, line 128: This should also add a proper version by using:
```golang
Versioned: cbor.NewVersioned(entity.LatestDescriptorVersion),
```

cmd/account/entity.go
```diff
@@ -0,0 +1,138 @@
...
+		Nodes: []signature.PublicKey{
+			signature.PublicKey(ed25519.NewPublicKey("BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB=")),
+		},
+	}
+	entityJson, err := json.Marshal(entity)
```
kostko, line 133: This should probably use `json.MarshalIndent` for nicer output.

cmd/paratime/register.go
```diff
@@ -0,0 +1,69 @@
...
+		common.BroadcastOrExportTransaction(ctx, npa.ParaTime, conn, sigTx, nil, nil)
+	},
+}
+
+func init() {
+	registerCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	registerCmd.Flags().AddFlagSet(common.TxFlags)
+
+	rtDesc := registry.Runtime{}
+	rtDescJson, err := json.Marshal(rtDesc)
```
kostko, line 66: This should probably use `json.MarshalIndent` for nicer output.

cmd/paratime/register.go
```diff
@@ -0,0 +1,69 @@
...
+
+		common.BroadcastOrExportTransaction(ctx, npa.ParaTime, conn, sigTx, nil, nil)
+	},
+}
+
+func init() {
+	registerCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	registerCmd.Flags().AddFlagSet(common.TxFlags)
+
+	rtDesc := registry.Runtime{}
```
kostko, line 65: This is a bad (and invalid) example. Maybe use a modified example from the network (or tests) for something better?

cmd/paratime/stats.go
```diff
@@ -180,9 +179,9 @@ func (s *runtimeStats) printStats() {
...
 
-var runtimeStatsCmd = &cobra.Command{
-	Use:   "runtime-stats [<start-height> [<end-height>]]",
-	Short: "Show runtime statistics",
+var statsCmd = &cobra.Command{
+	Use:   "stats [<start-height> [<end-height>]]",
+	Short: "Show ParaTime statistics",
 	Args:  cobra.MaximumNArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg := cliConfig.Global()
```
kostko, line 187: Unrelated, but `oasis paratime stats --no-paratime` crashes due to a nil pointer dereference. Should probably return an error instead (before derefercing `npa.ParaTime` below).

cmd/paratime/show.go
```diff
@@ -26,8 +26,8 @@ const (
 	blockLatest = "latest"
 )
 
-var blockCmd = &cobra.Command{
-	Use:   "block <number> [ <tx-index> | <tx-hash> ]",
+var showCmd = &cobra.Command{
```
kostko, line 29: So the idea is then to duplicate this thing for consensus blocks? But there is already a `oasis network show` which does something completely different? Will we need `oasis network show block ...` and then (for consistency) `oasis paratime show block ...`?

matevz: Actually I would merge all of those. So `oasis paratime show 12345` would show you the block round `12345`. If there will be some ambiguity, we could do `oasis paratime show block:12345` or `oasis paratime show paratime:12345`. Same for the `network`.

cmd/paratime/stats.go
kostko: Or just change the flags so `--no-paratime` is not even available.

cmd/paratime/show.go
kostko: Ok.

cmd/account/entity.go
matevz: This was removed now and will be part of the docs.

cmd/account/entity.go
matevz: This was removed now and will be part of the docs.

cmd/paratime/register.go
matevz: This was removed now and will be part of the docs.

cmd/paratime/stats.go
matevz: Fixed by checking, if paratime was provided. Omitting --no-paratime won't work, because they may be no ParaTimes configured at all.

cmd/common/transaction.go
```diff
@@ -97,6 +106,11 @@ func SignConsensusTransaction(
 		return nil, fmt.Errorf("account does not support signing consensus transactions")
 	}
 
+	txCfg := GetTransactionConfig()
+	if txCfg != nil && txCfg.Encrypted {
```
kostko, line 110: This is not needed as you can directly access `txEncrypted` here.

cmd/common/transaction.go
```diff
@@ -97,6 +102,10 @@ func SignConsensusTransaction(
 		return nil, fmt.Errorf("account does not support signing consensus transactions")
 	}
 
+	if txEncrypted {
+		cobra.CheckErr("--encrypted not supported for consensus transactions")
```
kostko, line 106:
```suggestion
		return nil, fmt.Errorf("--encrypted not supported for consensus transactions")
```

kostko (approved):

PR #82, opened 2023-05-29
fix(account): Fix delegate/undelegate ParaTime param

matevz:
Fixes regression introduced in 9ad0ac8dfc00f356470d0a042d61054d82958359 in `account delegate`/`undelegate` where we removed the `--paratime` and `--no-paratime` parameters. The default ParaTime was always considered then, which isn't supported yet, but will be in the future.

This PR reintroduces `--paratime` and `--no-paratime` parameters.

Before:

```
$ oasis account undelegate 20000000000 oasis1qpkl3vykn9mf4xcq9eevmey4ffrzf0ajtcpvd7sk
Unlock your account.
? Passphrase: 
Error: delegations within ParaTimes are not supported; use --no-paratime
$ oasis account undelegate 20000000000 oasis1qpkl3vykn9mf4xcq9eevmey4ffrzf0ajtcpvd7sk --no-paratime
Error: unknown flag: --no-paratime
Usage:
  oasis account undelegate <shares> <from> [flags]

Flags:
      --account string       explicitly set account to use
      --format string        transaction output format (for offline/unsigned modes) [json, cbor] (default "json")
      --gas-limit uint       override gas limit to use (disable estimation) (default 18446744073709551615)
      --gas-price string     override gas price to use
  -h, --help                 help for undelegate
      --network string       explicitly set network to use
      --nonce uint           override nonce to use (default 18446744073709551615)
      --offline              do not perform any operations requiring network access
  -o, --output-file string   output transaction into specified file instead of broadcasting
      --unsigned             do not sign transaction
  -y, --yes                  answer yes to all questions

Global Flags:
      --config string   config file to use
```

After:

```
$ oasis account undelegate 20000000000 oasis1qpkl3vykn9mf4xcq9eevmey4ffrzf0ajtcpvd7sk
Unlock your account.
? Passphrase: 
Error: delegations within ParaTimes are not supported; use --no-paratime
$ oasis account undelegate 20000000000 oasis1qpkl3vykn9mf4xcq9eevmey4ffrzf0ajtcpvd7sk --no-paratime
Unlock your account.
? Passphrase: 
You are about to sign the following transaction:
Method: staking.ReclaimEscrow
Body:
  From:   oasis1qpkl3vykn9mf4xcq9eevmey4ffrzf0ajtcpvd7sk
  Shares: 20000000000
Nonce:  0
Fee:
  Amount: 0.0 TEST
  Gas limit: 1275
  (gas price: 0.0 TEST per gas unit)

Network:  testnet
ParaTime: none (consensus layer)
Account:  oscar
? Sign this transaction? (y/N)
```

### DIFF
--- a/cmd/account/delegate.go
+++ b/cmd/account/delegate.go
@@ -67,6 +67,6 @@ var delegateCmd = &cobra.Command{
 }
 
 func init() {
-	delegateCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	delegateCmd.Flags().AddFlagSet(common.SelectorFlags)
 	delegateCmd.Flags().AddFlagSet(common.TxFlags)
 }
--- a/cmd/account/undelegate.go
+++ b/cmd/account/undelegate.go
@@ -68,6 +68,6 @@ var undelegateCmd = &cobra.Command{
 }
 
 func init() {
-	undelegateCmd.Flags().AddFlagSet(common.SelectorNAFlags)
+	undelegateCmd.Flags().AddFlagSet(common.SelectorFlags)
 	undelegateCmd.Flags().AddFlagSet(common.TxFlags)
 }


kostko (commented):
Delegations within paratimes will soon be supported though so we should think about how to support both modes properly to avoid weirdness later where the default behavior would change.

kostko (approved):

PR #89, opened 2023-06-07
feat(cmd/network): Also show node status

kostko:


cmd/network/show.go
```diff
@@ -91,6 +91,17 @@ var showCmd = &cobra.Command{
...
 				cobra.CheckErr(err)
+
+				if nodeStatus, err := registryConn.GetNodeStatus(ctx, idQuery); err == nil {
+					err = prettyPrint(nodeStatus)
+					cobra.CheckErr(err)
+				}
+				return
+			}
+
+			if nodeStatus, err := registryConn.GetNodeStatus(ctx, idQuery); err == nil {
```
matevz, line 102: Why not always calling GetNodeStatus() above before the GetNode() block and without returning.

matevz (approved):

PR #90, opened 2023-06-07
Add macOS releases

jberci:
This also changes the way makefiles print colorized output, similar to what was done in oasisprotocol/deoxysii-rust#22, for better compatibility with macOS.

macOS releases are universal binaries that work on either architecture, with whatever is the minimum supported system version for the go compiler (High Sierra on Intel, Big Sur on Apple silicon).

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -17,9 +17,12 @@ on:
 
 jobs:
   tests:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
     # NOTE: This name appears in GitHub's Checks API.
-    name: tests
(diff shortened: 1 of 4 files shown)


kostko (commented):
Nice!

Can you also change tests (the ci-test workflow) to run on macOS? Probably can do similar as [here](https://github.com/oasisprotocol/oasis-core-ledger/blob/master/.github/workflows/ci-tests.yml#L20-L25) with something like:
```yaml
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
    # NOTE: This name appears in GitHub's Checks API.
    name: tests-${{ matrix.os }}
    runs-on: ${{ matrix.os }}
```

kostko (approved):

PR #91, opened 2023-06-08
feat(cmd): Add support for governance operations

kostko:


cmd/account/governance.go
```diff
@@ -0,0 +1,193 @@
...
+var (
+	govCmd = &cobra.Command{
+		Use:     "governance",
+		Short:   "Governance operations",
+		Aliases: []string{"gov"},
+	}
+
+	govSubmitProposalCmd = &cobra.Command{
+		Use:   "submit-proposal",
+		Short: "Submit a governance proposal",
```
matevz, line 29: Since we're using term "submit" in the transaction command for broadcasting existing transaction, let's mention explicitly this is a *new* proposal.

```suggestion
		Short: "Submit new governance proposal",
```

cmd/account/governance.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+var (
+	govCmd = &cobra.Command{
+		Use:     "governance",
+		Short:   "Governance operations",
+		Aliases: []string{"gov"},
+	}
+
+	govSubmitProposalCmd = &cobra.Command{
```
matevz, line 27: While you're at it, I would also migrate `network governance-proposal` to `account governance show`.

Could you also add `account governance list` to list all opened proposals and proposers?

cmd/account/governance.go
```diff
@@ -0,0 +1,193 @@
...
+		Use:   "submit-proposal",
```
matevz, line 28: Could we simply call it *submit*?

```suggestion
		Use:   "submit",
```

matevz (changes requested):
Some refactoring suggestions.

cmd/account/governance.go
```diff
@@ -0,0 +1,193 @@
...
+			acc := common.LoadAccount(cfg, npa.AccountName)
+			sigTx, err := common.SignConsensusTransaction(ctx, npa, acc, conn, tx)
+			cobra.CheckErr(err)
+
+			common.BroadcastOrExportTransaction(ctx, npa.ParaTime, conn, sigTx, nil, nil)
+		},
+	}
+
+	govSubmitProposalCancelUpgradeCmd = &cobra.Command{
+		Use:   "cancel-upgrade <proposal-id>",
```
matevz, line 84: IMO we should move this to a level above, so something like this:

```
oasis account governance submit upgrade
oasis account governance cancel upgrade
```

Also, do we need to call it cancel upgrade proposal or could `cancel <proposal-id>` be universal for any governance proposal?

matevz (changes requested):

cmd/account/governance.go
kostko: This is because the transaction method is called "submit proposal". I changed the commands now to "create proposal".

cmd/account/governance.go
kostko: No, it is the other way around, these are different governance proposal types, currently:
* upgrade
* cancel upgrade

We may add others in the future. And you cannot cancel an arbitrary proposal, this is specifically for cancelling already scheduled upgrades.

cmd/account/governance.go
kostko: Added both show and list, please take a look.

cmd/account/governance.go
kostko: Now it is "create proposal", we could also have just "create" but then it is less clear what is being created (only though a proposal is currently the only thing that can be created).

matevz (changes requested):
Current example output for Testnet:

```
$ ./oasis acc gov show 7
Proposal passed, results: map[yes:50502292650718313]

Voted stake: 50502292650718313 (99.74%), total voting stake: 50632443796309264
Voted yes stake: 50502292650718313 (100.00%), voted stake: 50502292650718313, threshold: 68%

Validators voted:
oasis1qz2tg4hsatlxfaf8yut9gxgv8990ujaz4sldgmzx,<none>,10103361401282754 (19.95%)
oasis1qqv25adrld8jjquzxzg769689lgf9jxvwgjs8tha,<none>,10101056743564861 (19.95%)
oasis1qq2vzcvxn0js5unsch5me2xz4kr43vcasv0d5eq4,<none>,10099736164903599 (19.95%)
oasis1qz424yg28jqmgfq3xvly6ky64jqnmlylfc27d7cp,<none>,10099128300187338 (19.95%)
oasis1qr37y56g92chzvsew54kj7gu47cxyly7jytt5rm0,<none>,10099010040779761 (19.95%)

Validators not voted:
oasis1qrwncs459lauc77zw23efdn9dmfcp23cxv095l5z,GateOmega,36925612846659 (0.07%)
oasis1qq60zmsfca0gvmm3v8906pn5zqtt4ee2ssexramg,Validatrium,33743956828846 (0.07%)
oasis1qzjm0zwfg4egs9kk4d9rkujudzk8pjp5rvxyr3ag,Munay Network,11702422579124 (0.02%)
oasis1qpq97fm6lf87jzms9agd6z902nh7axtxvus6m352,LDV,10378819794784 (0.02%)
oasis1qpz97gfrvj5xzx8jx7x9zweeq0rcf2q6jg4a09qz,Stardust Staking,10086070309641 (0.02%)
oasis1qr2jxg57ch6p3787t2a8973v8zn8g82nxuex0773,Doorgod,9076788760953 (0.02%)
oasis1qz9x0zpja6n25hc5242k2e60l6a7ke2zsq9cqrqz,SerGo,5044216011403 (0.01%)
oasis1qqxxut9x74dutu587f9nj8787qz4dm0ueu05l88c,Princess Stake,2823252249537 (0.01%)
oasis1qq45am6gzaur2rxhk26av9qf7ryhgca6ecg28clu,Jr,2017758254984 (0.00%)
oasis1qzlzczsdme4scprjjh4h4vtljnmx3ag4vgpdnqln,Alexander (aka Bambarello) Validator,1593638024547 (0.00%)
oasis1qrq7hgvv26003hy89klcmy3mnedrmyd7lvf0k6qn,Perfect Stake,1403980986777 (0.00%)
oasis1qzwe6xywazp29tp20974wgxhdcpdf6yxfcn2jxvv,Simply Staking,1261098909212 (0.00%)
oasis1qphcvmsh6mw98vtg0cg4dvdsen5tm0g3e58eqpnp,Appload,1108526799775 (0.00%)
oasis1qpc66dgff4wrkj8kac4njrl2uvww3c9m5ycjwar2,<none>,1008643976345 (0.00%)
oasis1qpswaz4djukz0zanquyh2vswk59up22emysq5am9,StakeService,798000138962 (0.00%)
oasis1qr6a9jyxhvnxuq5kmtvntr0uv6mntvzvuyv57aqc,Orion.Money,480502898774 (0.00%)
oasis1qr4vsan850vmztuy9r2pex4fj4wxnmhvlgclg500,<none>,297535440854 (0.00%)
oasis1qqgvqelw8kmcd8k4cqypcsyajkl3gq6ppc4t34n2,AnkaStake,200251058019 (0.00%)
oasis1qzw6ljanmz8h4pjggemlgcysdnamaj830vlwu8c7,Mirror Frequency,200069721755 (0.00%)
```

I'm picky, but let's make printing more consistent with other commands:
- add `Network:  Testnet`
- add `Proposal ID: 7`
- then something like `Status: PASSED`
- followed by `Results:`
- `map[yes:50502292650718313]` should be
  ```  
     Yes: 50502292650718313
     No: 0
     Abstein: 0
  ```
- and then `=== VALIDATORS VOTED ===` with *enumerated* list of voters and the same for `=== VALIDATORS NOT VOTED ===`

cmd/account/governance/list.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+
+	"github.com/oasisprotocol/cli/cmd/common"
+	cliConfig "github.com/oasisprotocol/cli/config"
+	"github.com/oasisprotocol/cli/table"
+)
+
+var govListCmd = &cobra.Command{
+	Use:     "list",
```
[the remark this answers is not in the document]
kostko: It seems strange to have this in two places? Also it seems strange to have the list in one place and the rest of operations in another? Should we instead put everything under `network governance`? I like that I can do `oasis [...] gov --help` and get everything related to governance that I can do (showing proposals, voting, ...).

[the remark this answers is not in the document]
matevz: > Should we instead put everything under `network governance`?

Agree. That seems most feasible now.

matevz (approved):
LGTM

PR #134, opened 2023-09-25
cmd/wallet: More reserved addresses aliases and checks

matevz:
This PR:
- Migrates ResolveAddress() from Oasis SDK to Oasis CLI
- Adds literals for the pools and other reserved addresses into address resolving code
- Documents the usage of `pool:` literal
- Bumps usage of client-sdk/go to 0.7.2

Fixes https://github.com/oasisprotocol/oasis-sdk/issues/1508

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/oasisprotocol/deoxysii v0.0.0-20220228165953-2091330c22b7
 	github.com/oasisprotocol/metadata-registry-tools v0.0.0-20220406100644-7e9a2b991920
 	github.com/oasisprotocol/oasis-core/go v0.2300.9
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.7.1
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -6,6 +6,7 @@ toolchain go1.21.3
 
 // v1.5.0 has broken uint parsing, use a commit with the fix until the
 // the maintainers merge the PR: https://github.com/spf13/cast/pull/144
+replace github.com/oasisprotocol/oasis-sdk/client-sdk/go => ../oasis-sdk/client-sdk/go
```
kostko, line 9: This seems wrong.

matevz: Yeah, that's the debug commit. Will drop it when merging.

kostko (commented):
Will this be removed from the SDK?

Also could we have some tests for these, maybe just try to use them in `account show`?

matevz:
> Will this be removed from the SDK?

Yes, see https://github.com/oasisprotocol/oasis-sdk/pull/1507

kostko (approved):

PR #141, opened 2023-10-11
feat(cmd/network): Add network set-chain-context subcommand

kostko:


matevz (approved):
Can you also add a short section in the docs before set-default command: https://github.com/oasisprotocol/cli/blob/master/docs/network.md?

kostko:
Added.

PR #146, opened 2023-10-19
wallet: Add import-file command

### LINKED ISSUE #140: CLI docs for node operators
In preparation for 23.x, tidy up Oasis CLI docs for existing node operators:
- mention in `wallet import` how to import `ed25519-raw` type from the existing .pem file
- mention in `wallet list` that `ed25519-raw` is **the** type for node operators
- add to `network governance cast-vote` the standard NPA disclaimer

matevz:
This PR:
- adds the new `wallet import-file` command for importing the existing keypair from the PEM file. This helps existing node operators to vote on the upcoming network upgrade, because the governance operations were deprecated in `oasis-node` and they are forced to use the Oasis CLI
- refactors the `wallet` command into its own subfolder
- adds Network-Account selectors admonitions to the `network` command as well

Fixes #140.


cmd/wallet/remoteSigner.go
kostko, line 1: Make sure to rename this as `remote_signer.go`.

cmd/wallet/import_file.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/oasisprotocol/cli/cmd/common"
+	"github.com/oasisprotocol/cli/config"
+	"github.com/oasisprotocol/cli/wallet"
+	walletFile "github.com/oasisprotocol/cli/wallet/file"
+)
+
+var importFileCmd = &cobra.Command{
+	Use:   "import-file <file_name> <account_name>",
```
kostko, line 19:
```suggestion
	Use:   "import-file <file.pem> <account>",
```

kostko (approved):
Would be cool to also have an example/test that performs the import.

matevz:
> Would be cool to also have an example/test that performs the import.

Testing the interactive commands is tricky. We could add `--yes` for import-file and other wallet (import, create) commands as well, but I don't think it's a good idea to encourage people having unencrypted accounts in their wallet. Let's leave this for another PR.

kostko:
> but I don't think it's a good idea to encourage people having unencrypted accounts in their wallet

Yeah we could just use it for tests but not include it in docs.

Issue #154, opened 2023-11-10
Add `oasis account entity init`


matevz:
Implement replacement for `oasis-node registry entity init`.

PR #155, opened 2023-11-10
feat: Add account entity init command

matevz:
Implements #154 .

### DIFF
--- a/cmd/account/entity.go
+++ b/cmd/account/entity.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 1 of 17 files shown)


cmd/account/entity.go
```diff
@@ -17,11 +19,60 @@ import (
...
+
+			if entityOutputFile == "" {
+				fmt.Printf("%s\n", descriptorStr)
+			} else {
+				fout, err := os.Create(entityOutputFile)
+				cobra.CheckErr(err)
+				defer fout.Close()
+
+				_, err = fout.Write(descriptorStr)
+				cobra.CheckErr(err)
```
kostko, line 71: Could also use `os.WriteFile(entityOutputFile, descriptorStr, 0644)` instead.

kostko (approved):

PR #158, opened 2023-11-16
docs: Make serverdir/node path consistent with other docs

matevz:
`internal.sock` resides in the data dir, which is `/serverdir/node` in the rest of the "Run your node" docs. In the CLI docs, we just used `/serverdir/internal.sock`. This PR adds the `node` subdirectory to make it consistent.


kostko (approved):
Personally, I'm using `data` for the datadir instead of `node`.

matevz:
Followup PR #159

PR #159, opened 2023-11-17
docs: Replace serverdir/ -> node/ for consistency with other docs

matevz:
See https://github.com/oasisprotocol/docs/pull/588#discussion_r1396389809

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -57,7 +57,7 @@ autodetect the native token symbol and decimal places, the chain domain
 separation context and registered ParaTimes.
 
 ```
-oasis network add-local testnet_local unix:/serverdir_testnet/internal.sock
+oasis network add-local testnet_local unix:/node_testnet/data/internal.sock
 ```
 
 ## List Networks {#list}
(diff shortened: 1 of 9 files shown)


docs/network.md
```diff
@@ -57,7 +57,7 @@ autodetect the native token symbol and decimal places, the chain domain
 separation context and registered ParaTimes.
 
 ```
-oasis network add-local testnet_local unix:/serverdir_testnet/internal.sock
+oasis network add-local testnet_local unix:/node/data_testnet/internal.sock
```
kostko, line 60: I think it would make more sense to be `/node_testnet/data` instead?

gw0:
```suggestion
oasis network add-local testnet_local unix:/node_testnet/data/internal.sock
```
I also believe `/node_testnet/data` would be better. Because Testnet and Mainnet nodes use a different configuration (`etc/`), different runtimes (`runtimes/`), and different binary versions (`bin/`)...

kostko (approved):

gw0 (approved):

PR #201, opened 2024-03-01
governance: Add proposal content to governance show

amela:


cmd/network/governance/proposal.go
```diff
@@ -0,0 +1,66 @@
...
+				proposal.Content.Upgrade.PrettyPrint(ctx, "", os.Stdout)
+			case proposal.Content.CancelUpgrade != nil:
+				fmt.Println("Proposal type: Cancel Upgrade")
+				proposal.Content.CancelUpgrade.PrettyPrint(ctx, "", os.Stdout)
+			case proposal.Content.ChangeParameters != nil:
+				fmt.Println("Proposal type: Change Parameters")
+				proposal.Content.ChangeParameters.PrettyPrint(ctx, "", os.Stdout)
+			default:
+				fmt.Println("unknown")
+			}
```
kostko, line 59: I don't get this, why don't you simply do:
```go
proposal.Content.PrettyPrint(ctx, "", os.Stdount)
```

cmd/network/governance/proposal.go
kostko, line 1: Please run `make fmt`.

cmd/network/governance/proposal.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+
+	"github.com/oasisprotocol/cli/cmd/common"
+	cliConfig "github.com/oasisprotocol/cli/config"
+	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
+)
+
+var propShowCmd = &cobra.Command{
+	Use:     "proposal-show <proposal-id>",
```
kostko, line 19: This is strange, I would put this under the existing `show` not create a new subcommand?

cmd/network/governance/proposal.go
amela: So that it can write proposal type in human friendly language. I can change it if you want.

kostko: Doesn't the other one do that as well? It should write something like:
```
Upgrade:
  ...
```
And you can add something like `"Proposal content:"` before this and specify prefix as `"  "`.

cmd/network/governance/proposal.go
kostko: But it doesn't show the content?

cmd/network/governance/proposal.go
ptrus: You can also update the `PrettyPrint` function in oasis-core if you think it would be nicer to also print out the "Proposal type: " part.

cmd/network/governance/proposal.go
amela: Yes, the current command does not show the content. Ok, I'll add content to the existing command. :)

cmd/network/governance/show.go
```diff
@@ -254,8 +255,15 @@ var govShowCmd = &cobra.Command{
 			}
 		}
 
+		// Display the proposal content.
+
+		fmt.Println("=== PROPOSAL CONTENT ===")
+		proposal.Content.PrettyPrint(ctx, " ", os.Stdout)
```
matevz, line 261: It's a toplevel print, so we don't need any indent.

```suggestion
		proposal.Content.PrettyPrint(ctx, "", os.Stdout)
```

kostko (commented):
Should we modify the output so that by default it doesn't show the vote breakdown by validator but only summary statistics? Because listing validators on Mainnet makes the output quite long if you only want to see the proposal content.

Then if one wants to see vote details there could be an argument `--show-votes` or something.

matevz (changes requested):

amela:
> Should we modify the output so that by default it doesn't show the vote breakdown by validator but only summary statistics? Because listing validators on Mainnet makes the output quite long if you only want to see the proposal content.
> 
> Then if one wants to see vote details there could be an argument `--show-votes` or something.

I think this would be nice. WIll add it.

matevz:
> > Should we modify the output so that by default it doesn't show the vote breakdown by validator but only summary statistics? Because listing validators on Mainnet makes the output quite long if you only want to see the proposal content.
> > Then if one wants to see vote details there could be an argument `--show-votes` or something.
> 
> I think this would be nice. WIll add it.

Also, don't forget to update the document for this flag.

E.g. "You can also view individual validator votes by passing the `--show-votes` parameter:" and then add new `show-votes.in.static` and `show-votes.out.static` files and reference them.

amela:
Added `--show-votes` parameter

kostko (approved):

PR #206, opened 2024-03-06
Bump client-sdk to 0.8.1

kostko:
This also adds RPC endpoint migrations and improvements to support for different denominations.


cmd/account/transfer.go
```diff
@@ -17,15 +17,23 @@ import (
 )
 
 var transferCmd = &cobra.Command{
-	Use:     "transfer <amount> <to>",
+	Use:     "transfer <amount> [<denom>] <to>",
```
matevz, line 20: Interesting usage. I was thinking of introducing a slightly different flow for setting the validator's commission fee without the space character (e.g. `--bounds 329000/2%/4%`). A separate denominator argument makes sense here though. :+1:

matevz (changes requested):
Please also update the following files to reflect the new .io domain:
- examples/account/config/cli.toml
- examples/addressbook/config/cli.toml
- examples/network-set-rpc/01-set-rpc.in
- examples/network-set-rpc/config/cli.toml
- examples/network/add-tcpip.in.static
- examples/network/config/cli.toml
- examples/paratime-remove/config/cli.toml
- examples/wallet/config/cli.toml

kostko:
> Please also update the following files to reflect the new .io domain:

But then it won't test the migrations?

kostko:
Done, please take another look.

config/config.go
```diff
@@ -192,19 +192,54 @@ func (cfg *Config) migrateNetworks() bool {
...
+		for ptName, pt := range net.ParaTimes.All {
+			ptCfg, knownPt := defaultCfg.ParaTimes.All[ptName]
+			if !knownPt {
+				continue
+			}
+			if pt.ConsensusDenomination == ptCfg.ConsensusDenomination {
+				continue
+			}
+
+			pt.ConsensusDenomination = ptCfg.ConsensusDenomination
```
matevz, line 230: In the migrated config file I notcied the consensus denomination is set to `_`:

```ini
[networks.mainnet.denomination]
decimals = 9
symbol = 'ROSE'

[networks.mainnet.paratimes.sapphire]
consensus_denomination = '_'
```

Is this expected?

matevz (approved):
I tested the new denomination-specific operations and they work fine. However, I would suggest the following additions (can be in another PR):
- `oasis paratime list` should have another column `DENOMINATION(S)` containing comma-separated currencies, number of decimals and the default one marked with `(*)`. For example `EUROe[18] (*), TEST[18]`
- new subcommand `oasis paratime denom` with
  - `add <network> <paratime> <denomination> <number_of_decimals>`
  - `remove <network> <paratime> <denomination>`
  - in the future, other ERC-20 related operations will also be covered by the `denom` subcommand
- the list of registered denominations. This should probably go inside the `oasis paratime show <paratime>`

config/config.go
kostko: Yeah this is because the native denomination is represented by an empty string which is not handled correctly during config loading.

There should also be a section under paratime denominations named `_` representing the native denomination of the paratime.

PR #212, opened 2024-03-20
feat(cmd/network/governance): Improve governance proposal display

kostko:


cmd/network/governance/show.go
```diff
@@ -308,23 +308,34 @@ var (
 				cobra.CheckErr(fmt.Errorf("unexpected proposal state: %v", proposal.State))
 			}
 
+			fmt.Println()
+
+			// Display the proposal content.
+
+			fmt.Println("=== PROPOSAL CONTENT ===")
+			proposal.Content.PrettyPrint(ctx, "", os.Stdout)
```
matevz, line 316: I'm getting a bunch of parameters listed, but only one was actually changed. Perhaps you could fix the PrettyPrint as well to show only the relevant changes and hide the untouched parameters?

```
Change Parameters:
  Module: staking
  Changes: 
    - Parameter: min_delegation
      Value: <nil>
    - Parameter: reward_schedule
      Value: [map[scale:[1 27] until:90000]]
    - Parameter: min_transact_balance
      Value: <nil>
    - Parameter: fee_split_weight_vote
      Value: <nil>
    - Parameter: fee_split_weight_propose
      Value: <nil>
    - Parameter: reward_factor_epoch_signed
      Value: <nil>
    - Parameter: reward_factor_block_proposed
      Value: <nil>
    - Parameter: min_transfer
      Value: <nil>
    - Parameter: min_commission_rate
      Value: <nil>
    - Parameter: fee_split_weight_next_propose
      Value: <nil>
```

matevz (approved):
Don't forget to update 
- examples/network-governance/show.out.static
- examples/network-governance/show-votes.out.static.

cmd/network/governance/show.go
kostko: Unfortunately that pretty print implementation is in Core.

ptrus: Pretty print was already fixed in oasis-core, will be fixed once there's an oasis-core release with the fix included (and dep in cli is bumped).

Issue #106, opened 2023-07-11
Trim 0x when importing secp256k1 private key


matevz:
1. `oasis wallet import`
2. select raw private key
3. select sep256k1
4. paste private key
5. if the key starts with `0x` which is often in Eth world, CLI complains that it's not a valid hex value.

Expeceted behavior:
Automatically trim the leading `0x` characters, if they exist, and then import the private key without complaining.

Issue #213, opened 2024-03-26
`oasis wallet import` add non-interactive mode


matevz:
Currently, the only way to import an existing account is using the interactive mode. Sometimes you poses a bunch of private keys you would like to import and examine. In this case something similar to `oasis wallet create` would be useful. e.g.

```shell
oasis wallet import my_imported_raw_account --algorithm secp256k1-raw --secret 0x123456890abcdef1234569abcdef
oasis wallet import my_imported_mnemonic_account --algorithm secp256k1-bip44 --number 2 --secret "test test test test test test test test test junk"
```

The secret flag could also be omitted perhaps `oasis wallet import <name> [secret] [flags]`?

Related #120

PR #231, opened 2024-05-08
fix(cmd/wallet): Trim 0x when importing secp256k1

amela:
Fixes https://github.com/oasisprotocol/cli/issues/106

### DIFF
--- a/wallet/file/file.go
+++ b/wallet/file/file.go
@@ -279,9 +279,9 @@ func (af *fileAccountFactory) DataValidator(kind wallet.ImportKind, rawCfg map[s
 				}
 			case wallet.AlgorithmSecp256k1Raw:
 				// Ensure the private key is hex encoded.
-				_, err := hex.DecodeString(ans.(string))
+				_, err := Secp256k1FromHex(ans.(string))
 				if err != nil {
-					return fmt.Errorf("private key must be hex-encoded (without leading 0x): %w", err)
+					return fmt.Errorf("private key must be hex-encoded: %w", err)
 				}
 			case wallet.AlgorithmSr25519Raw:
 				// Ensure the private key is base64 encoded.


wallet/file/file.go
```diff
@@ -279,6 +280,7 @@ func (af *fileAccountFactory) DataValidator(kind wallet.ImportKind, rawCfg map[s
 				}
 			case wallet.AlgorithmSecp256k1Raw:
 				// Ensure the private key is hex encoded.
+				ans = af.DataTransformer(kind, rawCfg)(ans)
```
kostko, line 283: Why add this `DataTransformer` stuff when you could just do the two-line transformation here?

kostko (commented):
Also add an example to `examples/` so this code path is tested.

wallet/file/file.go
amela: Because validator does not transform the answer - it only validates. As I understand we want to trim 0x from the answer or not?

kostko: Ah didn't see that this is the validator function.

Looking now at `Import` itself, it seems that the `0x` prefix is already supported by `Secp256k1FromHex` that is used during the actual import? So instead of introducing `DataTransformer` just change this validation to use `Secp256k1FromHex` instead of `hex.DecodeString` and it should work? Also update the error message below.

examples/wallet/import-0x-secp256k1-raw.in.static
kostko, line 1: This shouldn't be static as otherwise this does not test the actual import. Oh I guess the problem is that in non-static mode you cannot answer the survey prompts? In this case remove the example as it is not doing anything.

kostko: cc @matevz I remember you were looking into how to handle this.

wallet/file/file.go
amela: Nice. Ok, will do it like this, but this will not trim the `0x`.

examples/wallet/import-0x-secp256k1-raw.in.static
matevz: There was no proper interactive way to test it at the time. We added `-y` flag to some commands to have at least some kind of testing there.

btw. The static examples are just for the documentation. And I agree that we don't need to document this. It's just convenient to silently trim 0x.

amela: But we are not trimming 0x anymore. Now we validate the input with `Secp256k1FromHex`

wallet/file/file.go
```diff
@@ -279,7 +279,7 @@ func (af *fileAccountFactory) DataValidator(kind wallet.ImportKind, rawCfg map[s
 				}
 			case wallet.AlgorithmSecp256k1Raw:
 				// Ensure the private key is hex encoded.
-				_, err := hex.DecodeString(ans.(string))
+				_, err := Secp256k1FromHex(ans.(string))
 				if err != nil {
 					return fmt.Errorf("private key must be hex-encoded (without leading 0x): %w", err)
```
kostko, line 284:
```suggestion
					return fmt.Errorf("private key must be hex-encoded: %w", err)
```

examples/wallet/import-0x-secp256k1-raw.in.static
matevz: Of course we do ;) https://github.com/oasisprotocol/cli/blob/f4e49b22888166d0296a783ae8d4ebad6e735e9e/wallet/file/secp256k1.go#L42

matevz: > There was no proper way to test interactive commands at the time. We added -y flag to some commands to have at least some kind of testing there.

FYI when https://github.com/oasisprotocol/cli/issues/213 is implemented, then we can properly test the wallet import commands.

amela: Hmm. As is understand this validator will not save the trimmed version of the input. It just validates.

kostko: True, but that doesn't matter as both variants are supported by the account loader.

kostko (approved):

PR #238, opened 2024-05-13
docs: Document --format parameter

matevz:
Documents the `--format` parameter.

Requires https://github.com/oasisprotocol/docs/pull/815

PR #240, opened 2024-05-15
docs: Line number requires specified language

matevz:
This is the behavior introduced by https://github.com/oasisprotocol/docs/pull/820. This is also the same as in the official markdown code block.

e.g. This doesn't work

````
```{2-3}
a
b
c
d
```
````

but this does:
````
```text {2-3}
a
b
c
d
```
````

### DIFF
--- a/docs/transaction.md
+++ b/docs/transaction.md
@@ -40,7 +40,7 @@ signature:
 
 ![code shell](../examples/transaction/show-invalid.in)
 
-![code {4}](../examples/transaction/show-invalid.out)
+![code text {4}](../examples/transaction/show-invalid.out)
 
 The `show` command is also compatible with ParaTime transactions. Take the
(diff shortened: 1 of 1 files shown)


lubej (approved):
Any reason why this changes aren't included in https://github.com/oasisprotocol/docs/pull/820?

kostko (approved):

matevz:
> Any reason why this changes aren't included in [oasisprotocol/docs#820](https://github.com/oasisprotocol/docs/pull/820)?

You probably meant https://github.com/oasisprotocol/cli/pull/238/files#diff-f47f6c25f8613e21081f4dd5cbbcdd07aaa87b7ebf8898b1ddfbde86cfedd1c5R43 ? Because the initial implementation of `code {4}` didn't actually work, but my tests used a cached version of the docs and thought it worked when merging it.

PR #243, opened 2024-05-21
Fix USB in macOS releases

jberci:


.goreleaser.yaml
```diff
@@ -5,24 +5,81 @@ before:
...
+      - "{{.Env.GOLDFLAGS_VERSION}}"
+  - id: oasis-darwin-arm64
+    binary: oasis
+    env:
+      - CC=oa64-clang
+      - CXX=oa64-clang++
+    goos:
+      - darwin
+    goarch:
+      - arm64
```
matevz, line 69: There is quite some duplicate code here. Did you consider using [github actions matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)?

matevz (changes requested):

.goreleaser.yaml
jberci: This should be a bit better now, I used yaml's map merge.

matevz: I suppose moving `build-common` outside of the `builds` doesn't work? Sth like:

```yaml
x-build-common: &build-common
  flags:
    - -trimpath
  ldflags:
    - -buildid=
    - "{{.Env.GOLDFLAGS_VERSION}}"
```

matevz (approved):
Try moving `build-common` outside of the `builds` category. Otherwise LGTM.

.goreleaser.yaml
jberci: It doesn't, unfortunately. I didn't find a way to declare a map without it also being part of the parsed result, i.e. with your example, goreleaser's struct would need a field called XBuildCommon or some such. You can only put anchors on actual data and then reference them.

kostko (approved):

Issue #247, opened 2024-06-03
Add `--subtract-fee` flag to include the gas fee in the amount


matevz:
Currently, there is no simple way to empty the account, because you will always need to leave some tokens to pay for the gas fee. Add `--subtract-fee` flag to transfer, withdraw, deposit commands which subtracts the required gas fee from the entered amount. By using this flag, you can check the account balance and then copy&paste it into the transfer command to empty it.

An alternative flag would be `--all`, but then the `amount` parameter would become optional which complicates things. Also, it allows you to completely empty the account and not leave, for example, some predetermined number of tokens in there.

The flag was inspired by the `subtractfeefromamount` argument used in bitcoin rpc: https://developer.bitcoin.org/reference/rpc/sendtoaddress.html

matevz:
Implemented.

PR #257, opened 2024-06-20
cmd: Add subtract-fee flag

matevz:
This PR decouples gas fee computation and signing the transactions. If `--subtract-fee` is provided, then the fee is subtracted from the transaction before signing it.

**Note**: Gas estimation for consensus transactions is higher for 8, because the nonce and the fee objects are now populated in the tx before estimating the gas.

Implements #247

### DIFF
--- a/cmd/account/account.go
+++ b/cmd/account/account.go
@@ -2,17 +2,27 @@ package account
 
 import (
 	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
 
 	"github.com/oasisprotocol/cli/cmd/account/show"
 )
 
-var Cmd = &cobra.Command{
-	Use:     "account",
-	Short:   "Account operations",
-	Aliases: []string{"a", "acc", "accounts"},
-}
+var (
+	subtractFee bool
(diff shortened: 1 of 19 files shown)


cmd/account/account.go
```diff
@@ -2,17 +2,27 @@ package account
...
 
 	"github.com/oasisprotocol/cli/cmd/account/show"
 )
 
-var Cmd = &cobra.Command{
-	Use:     "account",
-	Short:   "Account operations",
-	Aliases: []string{"a", "acc", "accounts"},
-}
+// SubtractFeeFlags is a force mode switch.
```
kostko, line 10: This comment seems to be in the wrong place (it covers the whole block) and also I have no idea what it means. Also, should the flag move under `common` so it could in theory be used outside `account`?

cmd/account/transfer.go
```diff
@@ -3,6 +3,7 @@ package account
 import (
 	"context"
 
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
```
kostko, line 6: This is in the wrong group, move below.

kostko:
Merged to fix examples.

PR #265, opened 2024-07-09
feat(cmd/rofl): Add initial support for ROFL app management

kostko:


cmd/rofl.go
```diff
@@ -0,0 +1,306 @@
...
+	roflAdminAddress string
+
+	roflCmd = &cobra.Command{
+		Use:     "rofl",
+		Short:   "ROFL app management",
+		Aliases: []string{"r"},
+	}
+
+	roflCreateCmd = &cobra.Command{
+		Use:   "create <policy.json>",
```
matevz, line 32: Please add an example policy to the docs.

cmd/rofl.go
```diff
@@ -0,0 +1,306 @@
...
+	// Parse policy.
+	var policy rofl.AppAuthPolicy
+	if err = json.Unmarshal(rawPolicy, &policy); err != nil {
+		cobra.CheckErr(fmt.Errorf("malformed ROFL app policy: %w", err))
+	}
+	return &policy
+}
+
+func init() {
+	f := flag.NewFlagSet("", flag.ContinueOnError)
```
matevz, line 282:
```suggestion
	adminFlags := flag.NewFlagSet("", flag.ContinueOnError)
```

matevz (approved):
Please also add a brief example to the docs (add `rofl.md`).

matevz:
How can a user know what's the staking threshold to register a ROFL app? This should probably go [somewhere here right](https://github.com/oasisprotocol/cli/blob/master/cmd/network/show.go#L361-L369)?

kostko:
Not there as that one is for network-wide stuff? The ROFL subcommand could mention it somewhere, e.g. when trying to create a new ROFL app and not having enough funds?

PR #283, opened 2024-09-10
feat(cmd/rofl): Add --mode argument to build command

### LINKED ISSUE #282: `rofl build`: Support unsafe builds for localnet
When you currently run `oasis rofl build` it executes `cargo build` with whichever environment variables set. This way you may accidentally build an unsafe version of your ROFL or the production version which won't run on your localnet.

I suggest we add sth like `--mode` with values: 
- `production` unsets any OASIS_UNSAFE flags when running child process,
- `unsafe` sets OASIS_UNSAFE flags and adds `--feature debug-mock-sgx`
- `auto` (default) detects the provided `NPA` and if the ParaTime is running in TEE, then use `production` otherwise (e.g. sapphire-localnet) use `unsafe`

kostko:
Fixes #282

### DIFF
--- a/build/cargo/cargo.go
+++ b/build/cargo/cargo.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os/exec"
+	"strings"
 )
 
 // Metadata is the cargo package metadata.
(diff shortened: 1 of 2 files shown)


build/cargo/cargo.go
```diff
@@ -48,14 +49,17 @@ func GetMetadata() (*Metadata, error) {
...
+func Build(release bool, target string, features []string) (string, error) {
 	args := []string{"build"}
 	if release {
 		args = append(args, "--release")
 	}
 	if target != "" {
 		args = append(args, "--target", target)
 	}
+	if features != nil {
+		args = append(args, "--feature", strings.Join(features, ","))
```
matevz, line 61:
```suggestion
		args = append(args, "--features", strings.Join(features, ","))
```

matevz (approved):
One glitch. Otherwise works perfectly, thanks!

PR #285, opened 2024-09-12
feat(rofl/build): Set x86_64-unknown-linux-gnu target for Localnet build

matevz:
On MacOS the default target must be set to x86_64 since this is the only platform we currently support.

Related https://github.com/oasisprotocol/sapphire-paratime/issues/377

### DIFF
--- a/cmd/rofl/build.go
+++ b/cmd/rofl/build.go
@@ -110,7 +110,7 @@ var (
 
 			// First build for the default target.
 			fmt.Println("Building ELF binary...")
-			elfPath, err := cargo.Build(true, "x86_64-unknown-linux-gnu", features)
+			elfPath, err := cargo.Build(true, "x86_64-unknown-linux-musl", features)
 			if err != nil {
 				cobra.CheckErr(fmt.Errorf("failed to build ELF binary: %w", err))
 			}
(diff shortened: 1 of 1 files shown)


kostko (approved):

cmd/rofl/build.go
```diff
@@ -71,14 +71,10 @@ var (
 			case buildMode == buildModeAuto && !offline:
 				ctx := context.Background()
 				conn, err := connection.Connect(ctx, npa.Network)
-				if err != nil {
-					break // Autodetection failed.
-				}
+				cobra.CheckErr(err)
```
kostko, line 74: This should produce a more verbose error message saying that autodetection failed, and to use `--mode` manually.

cmd/rofl/build.go
```diff
@@ -71,14 +71,10 @@ var (
...
 
 				params, err := conn.Consensus().Registry().ConsensusParameters(ctx, consensus.HeightLatest)
-				if err != nil {
-					break // Autodetection failed.
-				}
+				cobra.CheckErr(err)
```
kostko, line 77: This should produce a more verbose error message saying that autodetection failed, and to use `--mode` manually.

PR #298, opened 2024-10-03
feat(cmd/rofl): Add support for building TDX ROFL apps

kostko:


kostko:
It would be great if we could deterministically compile the ACPI tables, but my brief experiments with ASL did not help. So we are using per-CPU templates for now.

cmd/rofl/build/artifacts.go
```diff
@@ -0,0 +1,304 @@
...
+//
+// Returns the size of the created filesystem image in bytes.
+func createExt4Fs(fn, dir string) (int64, error) {
+	// Compute filesystem size in bytes.
+	fsSize, err := computeDirSize(dir)
+	if err != nil {
+		return 0, err
+	}
+	fsSize /= 1024                // Convert to kilobytes.
+	fsSize = (fsSize * 150) / 100 // Scale by overhead factor of 1.5.
```
abukosek, line 241: What is the typical size of the generated filesystems that you've tried so far?

Also, might be a good idea to add a check to make sure that `fn` is not an existing block device, so we don't accidentally trash a user's drive :)
(Currently, in the only place this function is called, the filename is hardcoded to `rootfs.ext4`, so it's not a problem, but it might be a problem in the future if we maybe have configurable filenames :man_shrugging:)

matevz (commented):
Following the existing tutorial and running `oasis rofl build tdx --mode unsafe` compiled the ROFL, but when it tried to pack it, I got

```
[built] path+file:///home/oa/rofl-oracle#0.1.0
Preparing stage 2 root filesystem...
Unpacking template...
Adding runtime as init...
Creating ext4 filesystem...
Error: failed to create rootfs image: exit status 1
```

Do I need to install additional dependencies? No other errors seem to be reported.

kostko:
Yes, let me add some checks whether the required tools exist.

matevz (approved):
Nice! I can confirm that this actually produces a working tdx orc file.

PR #308, opened 2024-10-16
feat(cmd/network): Add network show parameters command

### LINKED ISSUE #294: Add `network show parameters` command for showing current network parameters
You should be able to easily check the current consensus parameters such as the maximum size of the validator set etc. For starters we should cover those defined in the [genesis document](https://docs.oasis.io/node/genesis-doc).

abukosek:
Closes #294.


cmd/network/show.go
```diff
@@ -36,10 +38,11 @@ const (
 	selNativeToken
 	selGasCosts
 	selCommittees
+	selParameters
 )
 
 var showCmd = &cobra.Command{
-	Use:     "show { <id> | entities | nodes | paratimes | validators | native-token | gas-costs | committees }",
+	Use:     "show { <id> | entities | nodes | paratimes | validators | native-token | gas-costs | committees | parameters }",
```
matevz, line 45: I suggest we use alphabet order.

cmd/network/show.go
```diff
@@ -382,7 +390,90 @@ func showNativeToken(ctx context.Context, height int64, npa *common.NPASelection
...
+
+	doSection := func(name string, params interface{}) {
+		if outputFormat == formatJSON {
+			doc[name] = params
+		} else {
+			fmt.Printf("=== %s PARAMETERS ===\n", strings.ToUpper(name))
+			pp, err := json.MarshalIndent(params, "", "  ")
+			cobra.CheckErr(err)
+			out := string(pp)
+			out = strings.ReplaceAll(out, "{", "")
```
matevz, line 438: Move to PrettifyFromJSON helper?

matevz (commented):
Don't forget to update docs and examples.

examples/network-show/parameters.in.static
matevz, line 1: .static is used for outputs that should not be checked because of regular changes (e.g. nonces, balances etc.). Network parameters should remain fairly stable right? Did you try without .static suffix?

abukosek: Done :)

matevz (approved):
Looks good.

While you're at it, can you make these

https://github.com/oasisprotocol/cli/blob/master/docs/network.md?plain=1#L406-L411

consistent with your

```
By passing `--format json`, the output is formatted as JSON.
```

No need to dump this very long JSON in the docs.

PR #310, opened 2024-10-18
docs: reference Sapphire localnet consolidated page under tools

aefhm:
## Description

Necessary for https://github.com/oasisprotocol/docs/issues/959. Blocked by https://github.com/oasisprotocol/docs/pull/977

### DIFF
--- a/docs/rofl.md
+++ b/docs/rofl.md
@@ -64,7 +64,7 @@ The following SGX-specific flags are supported:
   Core runtime], otherwise the app may crash during the execution under actual
   SGX. It can be greater, if the app needs to spawn more threads.
 
-[Sapphire Localnet]: https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/guide.mdx#running-a-private-oasis-network-locally
+[Sapphire Localnet]: https://github.com/oasisprotocol/docs/blob/main/docs/dapp/tools/localnet.mdx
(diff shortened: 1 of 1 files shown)


docs/rofl.md
```diff
@@ -64,7 +64,7 @@ The following SGX-specific flags are supported:
   Core runtime], otherwise the app may crash during the execution under actual
   SGX. It can be greater, if the app needs to spawn more threads.
 
-[Sapphire Localnet]: https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/guide.mdx#running-a-private-oasis-network-locally
+[Sapphire Localnet]: ../../../dapp/tools/localnet.mdx
```
matevz, line 67: Since the localnet.mdx file is not part of the CLI repo, you need to use full github URL:

```suggestion
[Sapphire Localnet]: https://github.com/oasisprotocol/docs/blob/main/docs/dapp/tools/localnet.mdx
```

This way it's self-sustained if you browse CLI docs outside of docs.oasis.io and it will be mapped to the corresponding internal chapter, if viewed within docs.oasis.io.

aefhm: Ah, okay, so the trade off crossing repos is that we have to do URL reference and lose the relative links.

matevz (approved):

rube-de (approved):

PR #312, opened 2024-10-21
metadata: Bump oasisscan API to v2

amela:


matevz:
Are there any V2 API changes we need to be aware of?

amela:
@matevz afaik v2 has only one year of history. We use oasisscan's API only for:
`oasis network governance show <proposal-id> --show-votes`
I tested it with v2 and it still works.

matevz (approved):

PR #313, opened 2024-10-22
feat(cmd/network/status): Sort paratimes

amela:


cmd/network/status.go
```diff
@@ -132,7 +135,22 @@ var (
 					fmt.Println("==== ParaTimes ====")
 				}
 
-				for _, runtime := range nodeStatus.Runtimes {
+				keys := make([][]interface{}, 0, len(nodeStatus.Runtimes))
```
matevz, line 138: I would just store the map keys in the `keys` array. The sorting logic (ID vs name) should all come in the `sort.Slice()` function.

```suggestion
				keys := make([]string, 0, len(nodeStatus.Runtimes))
```

cmd/network/status.go
```diff
@@ -132,7 +135,40 @@ var (
...
+						jDescriptorID = jDescriptor.ID.String()
+						jParatimeName = getParatimeName(cfg, jDescriptorID)
+					}
+
+					switch iParatimeName == jParatimeName {
+					case true:
+						return iDescriptorID < jDescriptorID
+					default:
+						return iParatimeName < jParatimeName
+					}
```
matevz, line 167:
```suggestion
					var iParatimeName, jParaTimeName string

					iDescriptor := nodeStatus.Runtimes[keys[i]].Descriptor
					jDescriptor := nodeStatus.Runtimes[keys[j]].Descriptor

					if iDescriptor != nil {
						iDescriptorID := iDescriptor.ID.String()
						iParatimeName = getParatimeName(cfg, iDescriptorID)+iDescriptorID
					}

					if jDescriptor != nil {
						jDescriptorID := jDescriptor.ID.String()
						jParatimeName = getParatimeName(cfg, jDescriptorID)+jDescriptorID
					}

					return iParatimeName < jParatimeName
```

matevz (commented):
I think you can simply append the ParaTime ID to the name and you will get a deterministic order?

matevz (approved):

matevz:
Works as expected. Merging.

PR #319, opened 2024-11-15
cmd: Change rust target for building ROFL apps

jberci:
Some docker images have changed from being based on Alpine to Ubuntu.


kostko (approved):

matevz (commented):
Agree. Makes sense since we moved from alpine back to debian in sapphire-localnet and emerald-localnet images.

matevz (approved):

PR #332, opened 2024-12-09
feat(cmd/paratime): Print ROFL app id in rofl1 format

amela:
Fixes https://github.com/oasisprotocol/oasis-sdk/issues/2088. It looks like there is no need to add Pretty Print to `oasis-sdk`. The decoder only needs to be included.


matevz (approved):
Awesome, good job!

PR #333, opened 2024-12-10
cmd/paratime/show: Add show events command

ptrus:
Useful for easily fetching all events in a paratime block. The existing `paratime show` displays only block events (when querying info about a block) or events of a single transaction (when querying info about a transaction within a block).

Alternative would be updating the existing `paratime show <round>` to return all events emitted in a block, but the number of events could be large, so it might bloat the output for someone not really interested about this.

### DIFF
--- a/cmd/paratime/show.go
+++ b/cmd/paratime/show.go
@@ -18,6 +18,12 @@ import (
 	runtimeTx "github.com/oasisprotocol/oasis-core/go/runtime/transaction"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/accounts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/consensusaccounts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/contracts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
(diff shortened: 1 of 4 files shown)


cmd/paratime/show.go
```diff
@@ -35,15 +41,25 @@ const (
...
+}
+
 var (
 	outputFormat  string
 	selectedRound uint64
 
 	showCmd = &cobra.Command{
-		Use:     "show { <round> [ <tx-index> | <tx-hash> ] | parameters }",
-		Short:   "Show information about a ParaTime block, its transactions or other parameters",
+		Use:     "show { <round> [ <tx-index> | <tx-hash> ] | parameters | events }",
```
ptrus, line 61: These commands are currently inconsistent regarding the `round` argument.  `paratime show parameters` uses `--round` flag, while `paratme show <round>` uses `round` as a positional argument.

`show events` is consistent with `show parameters` since that was the easiest to implement. But this should probably be cleaned up (in a followup).

kostko: Initially it was not meant to show other things, so if we now have parameters and events, we should maybe also have `show block` or similar.

kostko (approved):

cmd/paratime/show.go
matevz: The initial idea was to resemble `oasis network show`. So try to autodetect what user wanted with as little typing as possible and then offer more explicit variants, if needed.

Some alternative ideas:
- `show <round> --events`
- `show events:<round>`
- `show <round>:events`
- or simply show all transactions with events when calling `show <round>`, also pass `--format json` and let user filter them with `jq`

IMO the current implementation makes the most sense.

ptrus: I'll merge it as is for now.

PR #335, opened 2024-12-12
feat(cmd/rofl): Add TDX container build support

kostko:


cmd/rofl/build/container.go
```diff
@@ -0,0 +1,177 @@
...
+
+	buildRofl "github.com/oasisprotocol/cli/build/rofl"
+	"github.com/oasisprotocol/cli/cmd/common"
+	roflCommon "github.com/oasisprotocol/cli/cmd/rofl/common"
+	cliConfig "github.com/oasisprotocol/cli/config"
+)
+
+const (
+	artifactContainerRuntime = "rofl-container runtime"
+	artifactContainerCompose = "docker-compose.yaml"
```
ptrus, line 26: Last time I tried it (maybe ~year ago), `podman-compose` (and even `podman` in general) wasn't really a drop-in replacement to `docker-compose` in practice, and it was somewhat tedious migrating things to `podman`. There were a lot of minor incompatibilities.

I hope this has improved, otherwise it might be better to be very clear that `podman` is used to run the containers. Maybe even go as far to use [compose.yaml](https://github.com/compose-spec/compose-spec/blob/main/spec.md#compose-file) filename, which seems to be preferred in the latest spec and avoid mentioning docker.

kostko: Good point, will rename it to `compose.yaml`.

cmd/rofl/mgmt.go
```diff
@@ -29,24 +33,150 @@ var (
...
 
-	createCmd = &cobra.Command{
-		Use:   "create <policy.yml>",
-		Short: "Create a new ROFL application",
+	appTEE         string
+	appKind        string
+	deploymentName string
+
+	initCmd = &cobra.Command{
+		Use:   "init <name> [--tee TEE] [--kind KIND]",
```
matevz, line 41: I would change behavior slightly:

```suggestion
		Use:   "init [PATH] [--tee TEE] [--kind KIND]",
```

So by default, it takes the current folder and creates rofl.yml with the project named after the folder name. If you provide `[PATH]`, then it creates the folder, names the project the same and uses it as a base folder for you rofl.yml.

matevz (commented):
I tested the `oasis rofl init` and it's really annoying that it registers your rofl app right away before you even start with the development. I would hardcode the app ID from the Localnet deployment as suggested in our docs: https://docs.oasis.io/rofl/deployment#checking-that-the-rofl-app-is-running

Then in the process of Testnet/Mainnet deployment, a proper app ID is obtained.

cmd/rofl/mgmt.go
kostko: This has been implemented.

cmd/rofl/build/build.go
```diff
@@ -54,13 +188,84 @@ func detectBuildMode(npa *common.NPASelection) {
...
+		ChainContext string               `json:"chain_context"`
+	}
+	root := trustRoot{
+		Height:       uint64(height),
+		Hash:         hash,
+		RuntimeID:    npa.ParaTime.Namespace(),
+		ChainContext: npa.Network.ChainContext,
+	}
+	encRoot := cbor.Marshal(root)
+	return base64.StdEncoding.EncodeToString(encRoot), nil
```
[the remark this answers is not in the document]
kostko: Root hash is not enough for a trust root.

> Would booting up the OS or building the container depend on which runtimeID/chain context of the trust root is being used

Yes, this is part of the enclave identity and is important to ensure integrity.

> Checking the root hash for debugging after .orc is built would also be easier because you would simply copy&paste it into the explorer and get the block right away

Can you explain what exactly would be easier? Do you mean just Base64 and CBOR decoding, so e.g. you would split everything into multiple env variables instead of having a single env variable? I don't see why this would be easier to debug, it is not like you will be manually inspecting ORC files and if you do then you can also decode Base64 and/or CBOR. And if you use a tool we can decode this for you and show you the underlying height/hash/...

cmd/rofl/build/build.go
```diff
@@ -21,13 +32,136 @@ const (
...
+			default:
+				fmt.Printf("unsupported TEE kind: %s\n", manifest.TEE)
+				return
+			}
+			if err != nil {
+				fmt.Printf("%s\n", err)
+				return
+			}
+
+			runScript(manifest, buildRofl.ScriptBuildPost)
```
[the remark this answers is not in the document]
kostko: Will add a separate `bundle-post` script that will have that information and will run after making the ORC.

cmd/rofl/mgmt.go
```diff
@@ -29,24 +33,150 @@ var (
...
+			deployment.AppID = appID.String()
+
+			fmt.Printf("Created ROFL application: %s\n", appID)
+
+			// Serialize manifest and write it to file.
+			data, _ := yaml.Marshal(manifest)
+			if err = os.WriteFile("rofl.yml", data, 0o644); err != nil { //nolint: gosec
+				cobra.CheckErr(fmt.Errorf("failed to write manifest: %w", err))
+			}
+		},
```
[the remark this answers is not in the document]
kostko: But that will then hide the obvious user error that there is no compose file defined and the resulting app will not do much?

build/rofl/manifest_test.go
```diff
@@ -0,0 +1,188 @@
...
+    cpus: 1
+    storage:
+        kind: ram
+        size: 16
+deployments:
+    default:
+        app_id: rofl1qpa9ydy3qmka3yrqzx0pxuvyfexf9mlh75hker5j
+        network: foo
+        paratime: bar
+`
```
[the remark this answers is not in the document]
kostko: Added.

cmd/rofl/mgmt.go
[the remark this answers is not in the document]
matevz: That's fine. There will be two files in the folder - `rofl.yml` and `compose.yml`. This way it will be obvious for the user what to do next and a good vibe already, because it compiles.

cmd/rofl/build/build.go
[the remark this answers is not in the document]
matevz: > Root hash is not enough for a trust root.

Can you elaborate? The block hash implicitly contains the chain context and the height (via previous block hashes).

> Can you explain what exactly would be easier? Do you mean just Base64 and CBOR decoding, so e.g. you would split everything into multiple env variables instead of having a single env variable? I don't see why this would be easier to debug, it is not like you will be manually inspecting ORC files and if you do then you can also decode Base64 and/or CBOR. And if you use a tool we can decode this for you and show you the underlying height/hash/...

Yes, Base64 and CBOR decoding is annoying. I assumed the root hash would be sufficient, so having just a single env variable.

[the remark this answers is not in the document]
kostko: You also need at least the height to make it easier to resolve as you need to start syncing the light client somewhere. And you do need the runtime ID, no way around it. I don't see the point of removing Base64/CBOR due to the above.

cmd/rofl/build/sgx.go
```diff
@@ -11,174 +11,129 @@ import (
...
+// sgxBuild builds an SGX-based "raw" ROFL app.
+func sgxBuild(
+	npa *common.NPASelection,
+	manifest *buildRofl.Manifest,
+	deployment *buildRofl.Deployment,
+	bnd *bundle.Bundle,
+) {
+	fmt.Println("Building an SGX-based Rust ROFL application...")
+
+	detectBuildMode(npa)
```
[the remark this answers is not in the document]
kostko: Done.

kostko:
This now also supports container secret management.

cmd/rofl/mgmt.go
kostko: Fixed the implementation so it behaves like `cargo init` et al. Please take another look.

matevz: Yeah, works exactly as expected now. Thanks.

matevz (commented):
May I suggest to use `.yaml` extension instead of `.yml` both for `rofl.yaml` and `compose.yaml` for consistency with the [Docker compose](https://docs.docker.com/compose/intro/compose-application-model/).

kostko:
Actually both `rofl.yaml` and `rofl.yml` work :) But yeah, we can change the default.

build/rofl/manifest.go
```diff
@@ -0,0 +1,312 @@
...
+	// AppID is the Bech32-encoded ROFL app ID.
+	AppID string `yaml:"app_id,omitempty" json:"app_id,omitempty"`
+	// Network is the identifier of the network to deploy to.
+	Network string `yaml:"network" json:"network"`
+	// ParaTime is the identifier of the paratime to deploy to.
+	ParaTime string `yaml:"paratime" json:"paratime"`
+	// Admin is the identifier of the admin account.
+	Admin string `yaml:"admin,omitempty" json:"admin,omitempty"`
+	// Debug is a flag denoting whether this is a debuggable deployment.
+	Debug bool `yaml:"debug,omitempty" json:"debug,omitempty"`
```
matevz, line 191: If people will be copying this for the Localnet, the debug should be visible in rofl.yaml, so they don't forget to change it.

```suggestion
	Debug bool `yaml:"debug" json:"debug,omitempty"`
```

cmd/rofl/build/build.go
```diff
@@ -2,65 +2,258 @@ package build
...
+
+			// Override the update manifest flag in case the policy does not exist.
+			if deployment.Policy == nil {
+				doUpdate = false
+			}
+
+			switch doUpdate {
+			case false:
+				// Ask the user to update the manifest manually.
+				fmt.Println("Update the manifest with the following identities to use the new app:")
```
matevz, line 158:
```suggestion
				fmt.Println("Update the manifest with the following identities under to use the new app:")
				fmt.Println("policy:")
				fmt.Println("  enclaves:")
				...
				fmt.Println("\nThen run oasis rofl update")
```

matevz (approved):
A few more nits, otherwise looks good.

cmd/rofl/mgmt.go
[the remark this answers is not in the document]
kostko: Added.

docs/rofl.md
```diff
@@ -22,7 +22,10 @@ The `build` command will execute a series of build commands depending on the
...
 Container (ORC) bundle.
 
-Building a ROFL bundle requires the [Network and ParaTime][npa] selectors.
+Building a ROFL bundle requires a ROFL app manifest (`rofl.yml`) to be present
+in the current working directory. All information about what kind of ROFL app
+to build is specified in the manifest.
+
 Additionally, the following flags are available:
 
 - `--mode` specifies a `production` (enabled SGX attestations suitable for the
```
matevz, line 31: This should also be removed now.

matevz: And `rofl.yml` -> `rofl.yaml` ;)

PR #336, opened 2024-12-12
feat: Unify command output format flag

### LINKED ISSUE #322: refactor: Move `--format` under common flags
The `--format` flag is used to change the output from human readable format (`text`) to `json` or `cbor`. Currently it has three implementations (account.go, notwork.go and paratime.go). Unify the behavior of this flag as a common flag and add it to specific commands that support it.

abukosek:
Closes #322.


matevz (approved):
Nice!

cmd/common/flags.go
```diff
@@ -3,26 +3,65 @@ package common
...
 
-// ForceFlag is a force mode switch.
-var ForceFlag *flag.FlagSet
+// Supported output formats for the format flag.
+const (
+	// Output plain text.
+	FormatText FormatType = "text"
+	// Output JSON.
+	FormatJSON FormatType = "json"
+)
```
matevz, line 36: I think we also use `cbor` when generating transactions.

abukosek: Yeah, I noticed that, but it seems to be only used by that specific command, so it didn't make sense to me to have it as a global option if no other command supports it :)

PR #342, opened 2025-01-07
ZigaMr/windows build support

ZigaMr:
This PR adds build-windows cross-compile option to Makefile and updates .goreleaser.yaml with windows build.


.github/workflows/ci-tests.yml
```diff
@@ -48,3 +48,40 @@ jobs:
           make clean-examples
           make examples
           git diff --exit-code examples
+
+  windows-tests:
```
matevz, line 52:
```suggestion
  tests-windows:
```

win_setup.ps1
matevz, line 1: This is in fact a Makefile version rewritten into a powershell script right? Should we simply call it `make.ps1`?

Or as an alternative, is there a downside, if we just use the existing Makefile and (GNU?) make running inside WSL for building the Oasis CLI on Windows?

scripts/gen_example.ps1
matevz, line 1: gen_example script generates docs snippets and can be just executed on Linux.

win_setup.ps1
ZigaMr: Yes, the powershell was added alongside Makefile since make isn't available in Windows natively, will rename to make.ps1.
This approach makes it easier to build since it does not require 3rd party apps (choco make) or WSL installation. 

If we just want to build a release version in the workflows (Ubuntu) it's possible to cross-compile for Windows with minimal changes to the Makefile, but this still means users will need to run WSL to compile locally.

scripts/gen_example.ps1
ZigaMr: Removed gen_example and win_setup, goreleaser now contains Windows build. Added option to cross-compile in Makefile.

Makefile
```diff
@@ -74,6 +74,10 @@ clean:
 	rm -f oasis
 	$(GO) clean -testcache
 
+windows-build:
```
matevz, line 77: Move this up next to `oasis`. I hope `find` works under WSL.

```suggestion
oasis.exe: $(shell find . -name "*.go" -type f) go.sum go.mod
```

Also add
```
build-windows: oasis.exe
```
 below `build`.

Makefile
```diff
@@ -83,3 +87,4 @@ clean:
 	$(lint-targets) lint \
 	$(test-targets) test \
 	clean
+	windows
```
matevz, line 90: Let's call this `build-windows` and move it up next to `build`.

matevz (approved):

matevz:
Please update the description of this PR to reflect the updated version.

Makefile
ZigaMr: Updated build-windows

Makefile
```diff
@@ -83,3 +88,4 @@ clean:
...
```
matevz, line 91: Missing backslash after `clean`.
`windows` should be called `build-windows` since that's the name of the rule.

Makefile
```diff
@@ -13,6 +13,11 @@ all: build
 
 # Build.
 build: oasis
+
+build-windows: $(shell find . -name "*.go" -type f) go.sum go.mod
```
matevz, line 17: This should be
```
build-windows: oasis.exe

oasis.exe: $(shell find . -name "*.go" -type f) go.sum go.mod
	@$(PRINT) "$(MAGENTA)*** Building for Windows...$(OFF)\n"
	GOOS=windows GOARCH=amd64 $(GO) build -v -o oasis.exe $(GOFLAGS) $(GO_EXTRA_FLAGS)
```

.github/workflows/ci-tests.yml
```diff
@@ -47,4 +47,4 @@ jobs:
         run: |
           make clean-examples
           make examples
-          git diff --exit-code examples
+          git diff --exit-code examples
```
matevz, line 50: Remove redundant newline.

PR #343, opened 2025-01-08
fix(cmd/network): Show ParaTimeID during node startup

amela:


abukosek (approved):

matevz (approved):
I can confirm this works all while initializing and being synced on Mainnet and Localnet. Good job.

PR #344, opened 2025-01-10
feat: Windows build support 2

ZigaMr:
This PR adds oasis.exe rule to Makefile.
Followup to https://github.com/oasisprotocol/cli/pull/342

PR #346, opened 2025-01-13
docs: Fix config path from Roaming to Local on Windows

matevz:
When testing #344 I realized the configuration directory is Local, not Roaming by default. Tested on Win10 64-bit.


ZigaMr:
Same on my machine (Win11, x64), saves to AppData/Local.

PR #351, opened 2025-01-23
feat(cmd/rofl): Perform early compose file validation

kostko:
This makes it faster to catch simple errors before deployment.


matevz (approved):
Nice! Works as expected.

PR #352, opened 2025-01-24
feat(cmd/rofl): Store artifacts on init for reproducible builds

kostko:
Also add `rofl artifacts-upgrade` for easy upgrades and bump default container runtime to rofl-containers v0.3.4.

### DIFF
--- a/cmd/rofl/mgmt.go
+++ b/cmd/rofl/mgmt.go
@@ -87,6 +87,9 @@ var (
 			height, err := common.GetActualHeight(ctx, conn.Consensus())
 			cobra.CheckErr(err)
 
+			blk, err := conn.Consensus().GetBlock(ctx, height)
+			cobra.CheckErr(err)
+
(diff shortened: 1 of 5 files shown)


cmd/rofl/mgmt.go
```diff
@@ -463,6 +479,37 @@ var (
 		},
 	}
 
+	artifactsUpgradeCmd = &cobra.Command{
+		Use:   "artifacts-upgrade",
```
matevz, line 483: We could simply use `upgrade` if we'll be updating some other components as well in the future (rofl.yaml file format, policy).

```suggestion
		Use:   "upgrade",
```

matevz (approved):

cmd/rofl/mgmt.go
kostko: My only concern is that we also have "update" and this sounds so similar?

PR #353, opened 2025-01-24
feat(cmd/rofl): Add --verify to build to verify enclave identities

kostko:
This makes it easier to audit apps.

### DIFF
--- a/cmd/rofl/build/build.go
+++ b/cmd/rofl/build/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"maps"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/rofl/build/build.go
```diff
@@ -147,6 +155,54 @@ var (
...
+					buildEnclaves[*eid] = struct{}{}
+				}
+
+				manifestEnclaves := make(map[sgx.EnclaveIdentity]struct{})
+				for _, eid := range deployment.Policy.Enclaves {
+					manifestEnclaves[eid] = struct{}{}
+				}
+
+				if !maps.Equal(buildEnclaves, manifestEnclaves) {
+					fmt.Println("Built enclave identities DIFFER from manifest enclave identities!")
```
matevz, line 171: Can you also print out the build enclave ID and the one that is in manifest file?

matevz (commented):
I tested this against my Localnet and it works fine.

While you're at it, could you also print the `Update the manifest with the following identities to use the new app:` and the new enclave IDs only, if they actually differ from the ones in the manifest file.

matevz (approved):

PR #391, opened 2025-03-11
feat(cmd/rofl): Only create deployments in create not init

kostko:
Fixes #385

PR #393, opened 2025-03-12
feat(cmd/rofl): Add repository/author/license and other app metadata

### LINKED ISSUE #375: rofl: Add repository, author, license fields, populate metadata
We should add common metadata fields like `repository`, `author`, `license` and others to the ROFL app manifest. Then we should use these fields to populate the per-app metadata that is already available. The metadata entries should be namespaced, e.g. with the `net.oasis.rofl.` or similar prefix.

kostko:
Based on #391 
Fixes #375

### DIFF
--- a/cmd/rofl/mgmt.go
+++ b/cmd/rofl/mgmt.go
@@ -66,6 +66,7 @@ var (
 			}
 
 			// Create a default manifest without any deployments.
+			// TODO: Extract author and repository from Git configuration if available.
 			manifest := buildRofl.Manifest{
 				Name:    appName,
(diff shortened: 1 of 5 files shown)


cmd/rofl/mgmt.go
```diff
@@ -66,6 +66,7 @@ var (
 			}
 
 			// Create a default manifest without any deployments.
+			// TODO: Extract author from Git configuration if available.
```
matevz, line 69: TODO: Extract author and git repository?

matevz (approved):

PR #405, opened 2025-03-18
cmd/rofl/build: Update manifest by default

matevz:
Running `oasis rofl build` now updates the manifest by default. Pass `--no-update-manifest` if you really don't want to update it. Mark `--update-manifest` as deprecated until all the examples and demos are updated to avoid breaking compatibility.

### DIFF
--- a/cmd/rofl/build/build.go
+++ b/cmd/rofl/build/build.go
@@ -34,7 +34,7 @@ var (
 	outputFn       string
 	buildMode      string
 	offline        bool
-	doUpdate       bool
+	noUpdate       bool
 	doVerify       bool
 	deploymentName string
 
@@ -49,10 +49,6 @@ var (
 				NeedAppID: true,
 				NeedAdmin: false,
 			})
-
-			if doVerify && doUpdate {
-				cobra.CheckErr("only one of --verify and --update-manifest may be passed")
-			}
 
 			fmt.Println("Building a ROFL application...")
 			fmt.Printf("Deployment: %s\n", deploymentName)
@@ -218,11 +214,11 @@ var (
 
 			// Override the update manifest flag in case the policy does not exist.
 			if deployment.Policy == nil {
-				doUpdate = false
+				noUpdate = true
 			}
 
-			switch doUpdate {
-			case false:
+			switch noUpdate {
+			case true:
 				// Ask the user to update the manifest manually (if the manifest has changed).
 				if maps.Equal(buildEnclaves, manifestEnclaves) {
 					fmt.Println("Built enclave identities already match manifest enclave identities.")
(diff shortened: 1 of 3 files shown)


cmd/rofl/build/build.go
```diff
@@ -218,11 +214,11 @@ var (
 
 			// Override the update manifest flag in case the policy does not exist.
 			if deployment.Policy == nil {
-				doUpdate = false
+				noUpdate = false
```
kostko, line 217: Shouldn't this be `true`?

cmd/rofl/build/build.go
```diff
@@ -333,9 +327,14 @@ func init() {
 	buildFlags := flag.NewFlagSet("", flag.ContinueOnError)
 	buildFlags.BoolVar(&offline, "offline", false, "do not perform any operations requiring network access")
 	buildFlags.StringVar(&outputFn, "output", "", "output bundle filename")
-	buildFlags.BoolVar(&doUpdate, "update-manifest", false, "automatically update the manifest")
+	buildFlags.BoolVar(&noUpdate, "no-update-manifest", false, "do not update the manifest")
 	buildFlags.BoolVar(&doVerify, "verify", false, "verify build against manifest and on-chain state")
 	buildFlags.StringVar(&deploymentName, "deployment", buildRofl.DefaultDeploymentName, "deployment name")
 
+	// TODO: Remove when all the examples, demos and docs are updated.
+	dummy := true
```
kostko, line 335:
```suggestion
	var dummy bool
```

cmd/rofl/build/build.go
```diff
@@ -333,9 +327,14 @@ func init() {
...
+	buildFlags.BoolVar(&dummy, "update-manifest", true, "not update the manifest")
+	_ = buildFlags.MarkDeprecated("update-manifest", "the app manifest is updated by default. Pass --no-update-manifest to prevent updating it.")
```
kostko, line 337:
```suggestion
	_ = buildFlags.MarkDeprecated("update-manifest", "The app manifest is now updated by default. Pass --no-update-manifest to prevent updating it.")
```

cmd/rofl/build/build.go
matevz: Oops, yes it should.

cmd/rofl/build/build.go
matevz: Actually, Cobra formats a message for you ending with comma, so the message should start with lower case:

```
Flag --update-manifest has been deprecated, the app manifest is now updated by default. Pass --no-update-manifest to prevent updating it.
```

kostko (approved):

PR #408, opened 2025-03-19
cmd: Show default config path for --config global flag

anzoman:
Sometimes users could struggle to find where the config file is located (especially on MacOS). It is useful if we show that info as part of the help message for the global `--config` flag. In the future, it would also be nice to have something like `oasis config set/get/reset`.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
-	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


anzoman:
I have this one a second though and although it is nice to see the config path directly in the CLI help, it also makes the help output look different on different platforms.

As an alternative, we could implement https://github.com/oasisprotocol/cli/issues/412.

PR #414, opened 2025-03-27
feat(cmd/rofl): Add support for pushing ORCs into OCI repositories

kostko:


build/rofl/oci.go
```diff
@@ -0,0 +1,114 @@
+package rofl
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	oras "oras.land/oras-go/v2"
```
matevz, line 10: Will you need to add `oras.land` dependency to oasis-sdk too in order to fetch .orc files on the node? Or there you could simply use HTTP GET?

kostko: Yes, but the corresponding Rust package (as the scheduler is written in Rust).

PR #424, opened 2025-04-10
feat(cmd/network/status): Add network names to chain context

amela:


cmd/network/status.go
```diff
@@ -75,7 +75,7 @@ var statusCmd = &cobra.Command{
 				fmt.Printf("Version:              %s", consensus.Version.String())
 				fmt.Println()
 
-				fmt.Printf("Chain context:        %s", consensus.ChainContext)
+				fmt.Printf("Chain context:        %s (%s)", npa.PrettyPrintNetwork(), consensus.ChainContext)
```
matevz, line 78: Actually the idea of the issue was to lookup the network name based on the chain context here: https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/go/config/default.go

So if you're running a node locally and you name the network `localhost`, it should still print `mainnet`, if it's running on `mainnet`.

matevz (approved):

matevz (changes requested):
Sorry, wrong button. Please properly resolve the comment ^^

matevz (approved):

PR #442, opened 2025-05-05
cmd&build: Push ORCs to Oasis-managed ROFL registry by default

anzoman:
Previously, the CLI required configuring an OCI repository manually when using `oasis rofl deploy`. The changes here simplify this by introducing a default Oasis-managed OCI registry for pushing ROFL ORCs.

### DIFF
--- a/build/rofl/oci.go
+++ b/build/rofl/oci.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"net/http"
 	"os"
 	"path/filepath"
 	"slices"
@@ -26,6 +27,9 @@ const (
 	ociTypeOrcLayer    = "application/vnd.oasis.orc.layer.v1"
 	ociTypeOrcArtifact = "application/vnd.oasis.orc"
 )
+
+// DefaultOCIRegistry is the default OCI registry.
+const DefaultOCIRegistry = "rofl.sh"
 
 // PushBundleToOciRepository pushes an ORC bundle to the given remote OCI repository.
 //
@@ -118,11 +122,17 @@ func PushBundleToOciRepository(bundleFn, dst string) (string, hash.Hash, error) 
 	if err != nil {
 		return "", hash.Hash{}, fmt.Errorf("failed to init OCI credential store: %w", err)
 	}
-	repo.Client = &auth.Client{
+	client := &auth.Client{
 		Client:     retry.DefaultClient,
 		Cache:      auth.NewCache(),
 		Credential: credentials.Credential(creds),
 	}
+	if strings.Contains(dst, DefaultOCIRegistry) {
+		client.Header = http.Header{
+			"X-Rofl-Registry-Caller": []string{"cli"},
+		}
+	}
+	repo.Client = client
 
 	// Push to remote repository.
 	if _, err = oras.Copy(ctx, store, tag, repo, tag, oras.DefaultCopyOptions); err != nil {
--- a/cmd/rofl/deploy.go
+++ b/cmd/rofl/deploy.go
@@ -8,7 +8,9 @@ import (
 	"maps"
 	"os"
 	"sort"
+	"time"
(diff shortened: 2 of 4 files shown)


cmd/rofl/deploy.go
```diff
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"time"
+
+	"github.com/google/uuid"
```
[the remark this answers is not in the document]
anzoman: Updated and linting goes through now.

build/rofl/oci.go
```diff
@@ -27,6 +28,9 @@ const (
 	ociTypeOrcArtifact = "application/vnd.oasis.orc"
 )
 
+// DefaultOCIRegistry is the default OCI registry.
+const DefaultOCIRegistry = "rofl-orc-registry.prd.oasis.io"
```
matevz, line 32: ?

```suggestion
const DefaultOCIRegistry = "oci.oasis.io"
```

cmd/rofl/deploy.go
```diff
@@ -114,8 +116,10 @@ var (
...
 			// Push ORC to OCI repository.
 			if deployment.OCIRepository == "" {
-				// TODO: Support default OCI repository.
-				cobra.CheckErr(fmt.Sprintf("Missing OCI repository for deployment '%s'.", deploymentName))
+				// Use default OCI repository with random name and timestamp tag.
+				deployment.OCIRepository = fmt.Sprintf(
+					"%s/%s:%d", buildRofl.DefaultOCIRegistry, uuid.New(), time.Now().Unix(),
+				)
 			}
 			fmt.Printf("Pushing ROFL app to OCI repository '%s'...\n", deployment.OCIRepository)
```
matevz, line 125: Can you store the config file's `oci_repository` field at this point, if the orc was successfully stored so we avoid multiple uploads, if someone forgets to pick the offer or term.

matevz (approved):
Two more suggestions, otherwise lgtm.

build/rofl/oci.go
kostko: We also have some `rofl.*` domains?

anzoman: What about `registry.rofl.app`/`orc-registry.rofl.app`?
The `oci.oasis.io` feels too general.

gw0: `registry.rofl.app` sounds nice. Or is the long-term plan to use `rofl.app` strictly for ROFL apps with open ports? If yes, then `registry.oasis.io`/`orc-registry.oasis.io`/`rofl-registry.oasis.io`?

kostko: Yes, the plan is to use `rofl.app` for app domains. What about any of the other ones?

kostko: Some ideas:
* `rofl.sh` (short and similar to `ttl.sh`, but then we dedicate the entire domain; could also be `registry.rofl.sh`)
* `registry.rofl.build`

cmd/rofl/deploy.go
kostko: (You just need to run `manifest.Save()` after successful push, see below.)

kostko (approved):

cmd/rofl/deploy.go
```diff
@@ -4,9 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/google/uuid"
```
kostko, line 7: This seems duplicated now, remove this one.

kostko (approved):

build/rofl/oci.go
anzoman: Went forward with `rofl.sh`.

cmd/rofl/deploy.go
anzoman: Removed.

PR #447, opened 2025-05-06
rofl deploy: Set default provider address, detect term and take the first offer

### LINKED ISSUE #441: `rofl deploy`: Autodetect term
Currently, the default term is `month`. For playground, we should have hourly plans. In this case, one always needs to provide `--term hour`. Introduce a new (default) term `auto` which picks the longest term that exists as an offer. In our case, the hour plan will be picked without additional parameters.

matevz:
This PR simplifies flow for first-time developers building ROFL on Testnet:
- detects the term (month, day, year) based on the selected offer (#441)
- sets the default ROFL provider to `oasis1qp2ens0hsp7gh23wajxa4hpetkdek3swyyulyrmz`, if none provided (Oasis foundation-managed Testnet node)
- if no offer provided, take the first one instead of throwing the error
- adds `--show-offers` for listing offers and quit

Fixes #441

### DIFF
--- a/cmd/rofl/deploy.go
+++ b/cmd/rofl/deploy.go
@@ -1,11 +1,13 @@
 package rofl
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"maps"
 	"os"
+	"sort"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
@@ -27,12 +29,13 @@ import (
 )
 
 var (
-	deployProvider  string
-	deployOffer     string
-	deployMachine   string
-	deployTerm      string
-	deployTermCount uint64
-	deployForce     bool
+	deployProvider   string
+	deployOffer      string
+	deployMachine    string
+	deployTerm       string
+	deployTermCount  uint64
+	deployForce      bool
+	deployShowOffers bool
 
 	deployCmd = &cobra.Command{
 		Use:   "deploy",
@@ -106,6 +109,12 @@ var (
 
 			fmt.Printf("Using provider: %s (%s)\n", machine.Provider, providerAddr)
 
+			if deployShowOffers {
+				// Display all offers supported by the provider.
+				showProviderOffers(ctx, npa, conn, *providerAddr)
+				return
+			}
+
 			// Push ORC to OCI repository.
 			if deployment.OCIRepository == "" {
 				// TODO: Support default OCI repository.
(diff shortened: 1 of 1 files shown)


cmd/rofl/deploy.go
```diff
@@ -249,6 +251,29 @@ var (
 	}
 )
 
+// detectTerm returns the preferred (longest) period of the given offer.
+func detectTerm(offer *roflmarket.Offer) (term roflmarket.Term) {
+	if offer == nil || offer.Payment.Native == nil {
+		cobra.CheckErr(fmt.Errorf("unable do determine payment term for offer '%s'", offer.ID))
```
kostko, line 257:
```suggestion
		cobra.CheckErr(fmt.Errorf("unable to determine payment term for offer '%s'", offer.ID))
```

kostko: This will trigger a nil pointer dereference in case `offer == nil` and you want to get `offer.ID`.

cmd/rofl/deploy.go
```diff
@@ -160,7 +156,8 @@ var (
 				}
 				var offer *roflmarket.Offer
 				for _, of := range offers {
-					if of.Metadata[provider.SchedulerMetadataOfferKey] == machine.Offer {
+					if of.Metadata[provider.SchedulerMetadataOfferKey] == machine.Offer || machine.Offer == "" {
```
kostko, line 159: Since `offers` are not sorted, this will take a random offer? We should probably explicitly sort them before.

cmd/rofl/deploy.go
```diff
@@ -249,6 +251,29 @@ var (
...
+		// Custom deploy term.
+		term = roflCommon.ParseMachineTerm(deployTerm)
+		if _, ok := offer.Payment.Native.Terms[term]; !ok {
+			cobra.CheckErr(fmt.Errorf("term '%s' is not available for offer '%s'", deployTerm, offer.ID))
+		}
+		return
+	}
+
+	for t := range offer.Payment.Native.Terms {
+		if t > term {
```
kostko, line 270: This is comparing the internal term identifier? What if we add `TermDay` with id `4`?

cmd/rofl/deploy.go
matevz: I'll order them by ID and pick the most fresh one (largest ID).

cmd/rofl/deploy.go
matevz: This would require Oasis SDK changes. I'll add a remark for now.

kostko (approved):

PR #455, opened 2025-05-09
cmd/rofl/deploy: Fix --provider behaviour

matevz:
This PR:
- fixes regression introduced by https://github.com/oasisprotocol/cli/pull/447/commits/ccc43dbb6fd331ea7f41ab6f58b16ac7a8a9967d . You couldn't deploy a new version of ROFL on the existing machine, because --provider parameter wasn't empty anymore.
- uses the Testnet provider only if Testnet Sapphire ParaTime is selected
- fixes CLI complaining, if you wanted to deploy to an existing machine that matches the `--provider` name.

### DIFF
--- a/build/rofl/provider/defaults.go
+++ b/build/rofl/provider/defaults.go
@@ -1,8 +1,20 @@
 package provider
 
-// DefaultSchedulerApp contains the default scheduler app IDs for each network/paratime.
-var DefaultSchedulerApp = map[string]map[string]string{
-	"testnet": {
-		"sapphire": "rofl1qrqw99h0f7az3hwt2cl7yeew3wtz0fxunu7luyfg",
+import "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+
+// RoflServices contains per network/ParaTime default ROFL services.
+type RoflServices struct {
+	// Scheduler is a ROFL app ID of the preferred scheduler app.
+	Scheduler string
+
+	// Provider is the address of the preferred provider.
+	Provider string
+}
+
+// DefaultRoflServices contains default built-in ROFL services for all supported networks/ParaTimes.
+var DefaultRoflServices = map[string]RoflServices{
+	config.DefaultNetworks.All["testnet"].ParaTimes.All["sapphire"].ID: {
+		Scheduler: "rofl1qrqw99h0f7az3hwt2cl7yeew3wtz0fxunu7luyfg",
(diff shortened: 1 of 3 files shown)


cmd/rofl/deploy.go
```diff
@@ -30,6 +31,8 @@ import (
 	cliConfig "github.com/oasisprotocol/cli/config"
 )
 
+const DEFAULT_TESTNET_ROFL_PROVIDER = "oasis1qp2ens0hsp7gh23wajxa4hpetkdek3swyyulyrmz"
```
kostko, line 34: This should be handled via some sort of a defaults map, per network, per paratime. See how it is done for scheduler apps in provider defaults.

cmd/rofl/deploy.go
```diff
@@ -92,14 +95,18 @@ var (
...
+					if npa.ParaTime.ID == config.DefaultNetworks.All["testnet"].ParaTimes.All["sapphire"].ID {
+						deployProvider = DEFAULT_TESTNET_ROFL_PROVIDER
+					} else {
+						cobra.CheckErr(fmt.Sprintf("Provider not configured for deployment '%s' machine '%s'. Please specify --provider.", deploymentName, deployMachine))
+					}
 				}
 
 				machine.Provider = deployProvider
 			default:
 				// Already set, require the provider to be omitted.
```
kostko, line 107: Or equal.

cmd/rofl/deploy.go
kostko: Something like this:
https://github.com/oasisprotocol/cli/blob/fd6094946bfa1b19ac7fb401faf2ff40f4c091c3/build/rofl/provider/defaults.go#L3-L8

build/rofl/provider/defaults.go
```diff
@@ -1,8 +1,17 @@
...
+import "github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+
+// DefaultRoflServices contains ROFL default services per network/ParaTime.
+type DefaultRoflServices struct {
+	Scheduler string
+	Provider  string
+}
+
+// DefaultSchedulerApp contains the ROFL services defaults for specific networks/ParaTimes.
+var DefaultSchedulerApp = map[string]DefaultRoflServices{
```
kostko, line 12: This name should be changed as it is not just the scheduler apps anymore.

kostko (approved):

PR #488, opened 2025-06-05
cmd/rofl/build: Validate image platform and size

### LINKED ISSUE #416: `rofl build`: Warn, if .orc size and images to be fetched is greater than `resources.storage.size`
A common mistake is leaving the default [512 MB](https://github.com/oasisprotocol/cli/blob/51ec2b674f4eb54f979685f56188051a066ede42/cmd/rofl/mgmt.go#L82) storage size in `rofl.yaml` while `compose.yaml` will fetch images with bigger size in total.

1. Detect the size of images to be pulled during startup and warn the user that `resources.storage.size` will be too small to accommodate them. This is also a good check that the images are really publicly available to download.

2. Also, when checking the image size, also check, if the platform of the Docker image is `linux/amd64`. See https://github.com/oasisprotocol/oasis-sdk/issues/2160 bullet 5.

3. Also, try to fetch all images in compose file to make sure they are publicly accessible.

abukosek:
Closes #416.

### DIFF
--- a/cmd/rofl/build/container.go
+++ b/cmd/rofl/build/container.go
@@ -2,12 +2,16 @@ package build
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
 	"golang.org/x/net/idna"
 
-	"github.com/compose-spec/compose-go/v2/cli"
+	compose "github.com/compose-spec/compose-go/v2/cli"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/registry/remote"
 
 	"github.com/oasisprotocol/oasis-core/go/runtime/bundle"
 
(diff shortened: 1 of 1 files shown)


matevz:
> I've multiplied the final size by 2, which seems like a good estimate for the commonly-used images 🤷‍♂️

I can confirm that `docker.io/ollama/ollama:latest` is ~1.6GB compressed and ~ 3.4GB uncompressed so it's a reasonable factor.

cmd/rofl/build/container.go
```diff
@@ -67,6 +76,86 @@ func tdxBuildContainer(
...
+			if platform.Architecture != "amd64" || platform.OS != "linux" {
+				return fmt.Errorf("compose file validation failed: image '%s' has incorrect platform (expected linux/amd64, got %s/%s)", imageFull, platform.OS, platform.Architecture)
+			}
+		}
+		platform = mf.Config.Platform
+		if platform != nil {
+			if platform.Architecture != "amd64" || platform.OS != "linux" {
+				return fmt.Errorf("compose file validation failed: image '%s' has incorrect platform (expected linux/amd64, got %s/%s)", imageFull, platform.OS, platform.Architecture)
+			}
+		}
```
matevz, line 137: Can't you deduplicate this code with
```go
for platform := range map[string]string{mainDescriptor, mf.Config} {
...
}
```
?

matevz (approved):

PR #510, opened 2025-06-27
cmd/account/deposit: Set gas price to 0 by default

matevz:
ParaTime deposit transactions don't require gas. Set it to 0 by default instead of autodetecting it.


matevz:
> I don't think I like this as this exemption should eventually go away.

Is there a minimum deposit amount defined on the network? That would prevent spam transactions then, right?

PR #512, opened 2025-06-28
cmd/rofl/logs: Add -y for non-interactive mode

matevz:
You typically don't encrypt test accounts when developing ROFLs, so the `-y` flag comes handy in this case to check the logs.


kostko (approved):

kostko:
Also we could cache the JWT to avoid the need to relogin.

PR #517, opened 2025-07-10
cmd/rofl: Populate compose.yaml with boilerplate content

matevz:
Based on the feedback from one of our partners, it would be a good idea to populate `compose.yaml` with ROFL-specific features (storage, appd, secrets) and then let developer remove those, if not needed.

### DIFF
--- a/cmd/rofl/init_artifacts/compose.yaml
+++ b/cmd/rofl/init_artifacts/compose.yaml
@@ -1,0 +1,14 @@
+services:
+  mycontainer:
+    build: .
+    image: "docker.io/org_name/image_name"      # Always provide FQDN image names!
+    platform: linux/amd64
+    volumes:
+      - my-volume:/root/.my-volume              # Persistent per-machine storage.
+      - /run/rofl-appd.sock:/run/rofl-appd.sock # appd REST API.
+    environment:
+      - MY_SECRET_1=${MY_SECRET_1}              # See `oasis rofl secret` command.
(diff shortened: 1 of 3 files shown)


cmd/rofl/init_artifacts/compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  mycontainer:
+    build: .
+    image: "docker.io/org_name/image_name" # Always provide FQDN image names!
+    platform: linux/amd64
+    volumes:
+      - /storage/my_container:/root/.my_volume  # Persistent per-machine storage.
```
kostko, line 7: This is not the best. Please see how to declare volumes correctly, eg. you should not use absolute paths but rather declare volumes explicitly.

Looking at docs it seems someone changed this as I remember I wrote a section on how to do this correctly.

kostko: I see docs about persistent volumes were changed in https://github.com/oasisprotocol/oasis-sdk/commit/c0c6a5bf765fd5775dd577bd558a66c3968c1dd6 and I didn't notice that part.

We should change it back to use volumes without referencing `/storage` directly IMO? At least that seems the most correct approach to me.

kostko: image

matevz: I agree. This approach is also cleaner because it uses preconfigured `/storage` path and it can be recycled in other docker images.

PR #543, opened 2025-07-16
cmd/rofl: Fix git repository initialization

### LINKED ISSUE #535: `rofl init` don't create git repo, if it already exists
If your project is already part for a git repo (folder above), it shouldn't create another `.git` inside it.

abukosek:
Fixes #535.


cmd/rofl/mgmt.go
```diff
@@ -117,21 +117,60 @@ var (
...
+						fmt.Printf("Git repository initialized.\n")
+					}
+				}
+			} else {
+				// Git repository already exists.
+				gitPath = curPath
 			}
 
-			// Initialize .gitignore.
+			// Initialize .gitignore in the appropriate path.
```
matevz, line 171: It's fine if `.gitignore` lives in the path where the ROFL app is and not the root.

cmd/rofl/mgmt.go
```diff
@@ -117,21 +117,60 @@ var (
...
 			}
 
-			// Initialize git repository if not already present.
-			_, err = os.Stat(".git")
-			if err != nil && errors.Is(err, os.ErrNotExist) {
-				// Git directory doesn't exist, try to initialize.
-				gitInitCmd := exec.Command("git", "init")
-				if err = gitInitCmd.Run(); err != nil {
-					fmt.Printf("Git repository not initialized: %s.\n", err)
+			// Save current working directory.
```
matevz, line 120: Since you rely on the `git` command for initializing the repo, I see no reason why not simply execute `git rev-parse` to check whether the project folder is already part of the git repo and if not, run `init`?

abukosek: That was my initial idea, but then realized that we need to modify the .gitignore file and need its path anyway.

kostko: Btw, we already have a "get me the Git root" as we need it when determining the project base dir for use with the Docker builder. Please take a look at `build/env/base.go` which should probably be used instead of reimplementing it.

cmd/rofl/mgmt.go
abukosek: OK, I've updated the code, please have another look :)

matevz (approved):

Issue #326, opened 2024-12-03
cli: Add Oasis CLI to brew


matevz:
@ahmedhamedaly suggested half of hackathoners used Mac for developing ROFL. Oasis CLI should be added to brew.

uniyalabhishek:
@matevz can you please help create a new repo named oasisprotocol/homebrew-oasis 

## Which path should we take?

| Path                                         | When to use                             | Installation command for users                         | Review burden                 | Time to ship |
| -------------------------------------------- | --------------------------------------- | ------------------------------------------------------ | ----------------------------- | ------------ |
| **Own tap** (`oasisprotocol/homebrew-oasis`) | Need it to battle test and in short-term for hackathons | `brew tap oasisprotocol/oasis`<br>`brew install oasis` | ✦ none (you control the repo) | **< 1 day**  |
| **homebrew‑core**                            | Long‑term discoverability               | `brew install oasis` (no tap step)                     | Homebrew maintainers          | 1‑4 weeks    |

▶ **Recommendation:** We ship the **tap now**, upstream to **core** once the formula is battle tested.
Nothing is thrown away, GoReleaser will generate (and later update) the very same Ruby file for both cases.

njelich:
Looking at the official requirements for inclusion in homebrew core:

[...]

**We don’t like tools that upgrade themselves**

Software that can upgrade itself does not integrate well with Homebrew formulae’s own upgrade functionality. The **self-update functionality should be disabled** (while minimising complication to the formula). It’s fine (and well-supported) for Casks.

[...]

The software in question must:

- be maintained (i.e. the last release wasn’t ages ago, it works without patching on all Homebrew-supported OS versions and has no outstanding, unpatched security vulnerabilities)
- be stable (e.g. not declared “unstable” or “beta” by upstream)
- **be known (e.g. GitHub repositories should have >=30 forks, >=30 watchers or >=75 stars)**
- be used
- have a homepage 

We will reject formulae that seem too obscure, partly because they won’t get maintained and partly because we have to draw the line somewhere.

**We frown on authors submitting their own work unless it is very popular.**

[...]

As such it would make sense to communicate clearly that the stars for a broader set of repos should be looked at, since the CLI is one of a bunch.

To make that easier, I have prepared the following graphic with an overview of stars for some of the bigger Oasis repos:

<img width="3664" height="2616" alt="Image" src="https://github.com/user-attachments/assets/388f4a91-7334-48df-91a0-f571a49fc19c" />

[...]

PR #546, opened 2025-07-18
feat: add brew tap for oasis-cli

uniyalabhishek:
Resolves https://github.com/oasisprotocol/cli/issues/326

this PR adds GoReleaser GitHub Action that builds and publishes tagged Oasis‑CLI releases. creates a Homebrew cask (Casks/oasis-cli.rb) in this repo, turning it into a tap.
the workflow auto‑updates the cask with each release and hosts macOS‑universal binaries. Mac devs can now 
`brew install --cask oasisprotocol/cli/oasis-cli`

(and later `brew upgrade --cask oasis-cli` to update).

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,3 +1,4 @@
+version: 2
 project_name: Oasis CLI
 
 before:
@@ -14,6 +15,7 @@ universal_binaries:
 
 env:
   - CGO_ENABLED=1
+  - GOLDFLAGS_VERSION=-X github.com/oasisprotocol/cli/version.Software={{ .Version }}
 
 builds:
   - &build-common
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -121,3 +124,18 @@ release:
...
+homebrew_casks:
+  - name: oasis-cli
+    binary: oasis
+    directory: Casks
+    homepage: "https://github.com/oasisprotocol/cli"
+    description: "Command‑line interface for the Oasis Network"
+    commit_author:
+      name: github-actions[bot]
+      email: github-actions[bot]@users.noreply.github.com
+    commit_msg_template: "chore(brew): update {{ .ProjectName }} to {{ .Tag }}"
```
matevz, line 137: I'm not familiar with the `homebrew_casks` workflow. But making a commit directly to the main branch will not work due to the security policies. Why would you need to commit to the cli repo?

.goreleaser.yaml
```diff
@@ -13,7 +14,8 @@ universal_binaries:
     name_template: oasis
 
 env:
-  - CGO_ENABLED=1
+  - CGO_ENABLED=0
```
matevz, line 17: This should remain the same.

```suggestion
  - CGO_ENABLED=1
```

.goreleaser.yaml
uniyalabhishek: created a fix to open PRs instead trying to commit directly. more info on goreleaser `homebrew_casks` https://goreleaser.com/customization/homebrew_casks/

we need to commit to the cli repo (now with a PR) to let goreleaser auto‑add the updated `Casks/oasis-cli.rb` with each tag so brew users can brew upgrade oasis-cli, without us maintaining a separate repo/tap

matevz: Could this simply still go under the `cli` repo but to a separate branch? Similarly to how `gh-pages` branch is commonly used for the github pages.

uniyalabhishek: after searching a bunch, this doesn’t looks to be possible. Homebrew only clones the default branch ( origin/HEAD ) of a tap

PR #547, opened 2025-07-21
feat(cmd/account): Show Eth addr and name if avaliable

amela:


cmd/account/show/show.go
```diff
@@ -44,6 +44,9 @@ var (
 			case npa.Account != nil:
 				// Default account is selected.
 				targetAddress = npa.Account.Address
+				if npa.AccountName != "" {
+					fmt.Printf("Name:             %s\n", npa.AccountName)
+				}
```
matevz, line 49: Printing the name has nothing to do with the selected account. The address should be looked up in the corresponding records (wallet config, addressbook, hardcoded addresses etc.) and printed if found. Check out how we did it for `accounts transfer`.

cmd/common/helpers.go
```diff
@@ -47,3 +47,26 @@ func GenAccountNames() types.AccountNames {
...
+		}
+	}
+
+	for name, acc := range cfg.AddressBook.All {
+		if acc.Address == address {
+			return name
+		}
+	}
+	return ""
+}
```
matevz, line 72: Why not just?

```suggestion
// FindAccountName finds account's name (if exists).
func FindAccountName(address string) string {
	an := GenAccountNames()
	return an[address]
}
```

matevz (commented):
One more suggestion, otherwise looks good.

PR #565, opened 2025-08-08
feat: add ability to disable update command for managed installs

njelich:
adds support for disabling the update command, useful for managed install such as the upcoming homebrew release

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -59,7 +59,7 @@ var (
 	githubAPIBase = "https://api.github.com"
 	download      = doDownload
 	osExecutable  = os.Executable
-	disableUpdateSubcommand = false
+	disableUpdateSubcommand = "false"
 )
 
 // semverDiffers reports whether versions a and b differ (ignoring leading 'v').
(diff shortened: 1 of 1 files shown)


cmd/update.go
```diff
@@ -59,6 +59,7 @@ var (
 	githubAPIBase = "https://api.github.com"
 	download      = doDownload
 	osExecutable  = os.Executable
+	disableUpdateSubcommand = "false"
```
matevz, line 62: Maybe just

```suggestion
	disableUpdateCmd = "false"
```

Also, are you allowed to use `-X` on lower-cased variables?

cmd/update.go
```diff
@@ -505,6 +506,10 @@ then downloads and replaces the current binary.`,
 }
 
 func init() {
+	// I have to use a string here because of the limitations of the -X flag in Go build.
```
matevz, line 509: I'd put this comment where you defined the variable instead.

cmd/update.go
njelich: Seems to work

cmd/update.go
```diff
@@ -59,6 +59,8 @@ var (
 	githubAPIBase = "https://api.github.com"
 	download      = doDownload
 	osExecutable  = os.Executable
+	// I have to use a string for DisableUpdateCmd because of the limitations of the -X flag in Go build.
```
matevz, line 62:
```suggestion
	// DisableUpdateCmd removes `oasis update` command. This is suitable for environments where the package manager handles updates.
```

Issue #433, opened 2025-04-23
`rofl secret set`: Add support for .env files


matevz:
Typically people store their secrets in .env file (`key=value` pairs with `#` for comments). For example `.env.production` or `.env.development`. Symlinks are also common for example `.env` &rightarrow; `.env.production`. In our case, one could also name .env files as `.env.localnet`, `.env.testnet` and `.env.mainnet`.

Add a new `oasis rofl secret import <dot-env-file>` command for importing secrets from .env files.

matevz:
Empyreal team uses `empctl secrets upsert --file secrets.json`

PR #571, opened 2025-08-13
feat: add rofl secret import cmd

uniyalabhishek:
resolves https://github.com/oasisprotocol/cli/issues/433

### DIFF
--- a/build/dotenv/dotenv.go
+++ b/build/dotenv/dotenv.go
@@ -1,0 +1,194 @@
+package dotenv
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+)
+
+// errUnterminatedQuote is returned when a quoted value is not properly terminated.
+var errUnterminatedQuote = errors.New("unterminated quoted value")
+
+// Parse parses a .env formatted string into key/value pairs.
+//
+// Supports lines like:
+//
+//	KEY=value
+//	KEY="quoted # not comment"
+//	KEY='single quoted'
+//	export KEY=value
+//
+// Comments start with '#' when unquoted (at start or preceded by whitespace).
+func Parse(data string) (map[string]string, error) {
+	res := make(map[string]string)
+
+	// Normalize line endings.
+	data = strings.ReplaceAll(data, "\r\n", "\n")
+	data = strings.ReplaceAll(data, "\r", "\n")
+
+	lines := strings.Split(data, "\n")
+	for i := 0; i < len(lines); i++ {
+		raw := lines[i]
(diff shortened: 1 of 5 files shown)


cmd/rofl/dotenv.go
kostko, line 1: I would move this to its own `build/dotenv` package.

cmd/rofl/dotenv.go
```diff
@@ -0,0 +1,190 @@
...
+	consumed := 0
+
+	for {
+		v, e := parseQuotedValue(combined, delim, unescape)
+		if e == nil {
+			return v, consumed, nil
+		}
+
+		// Only continue if this is the unterminated-quote condition.
+		if !strings.HasPrefix(e.Error(), "unterminated quoted value:") {
```
kostko, line 129: Given that you are generating these errors below, just create a new error type using `errors.New` and then match on it using `errors.Is` instead of matching strings.

And I would rewrite the above using a switch statement, e.g.:
```go
switch {
case err == nil:
    return v, consumed, nil
case errors.Is(err, errUnterminatedQuote):
    // ...
default:
    return "", 0, fmt.Errorf("line %d: %w", startIdx+1, err)
}
```

cmd/rofl/dotenv.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if strings.HasPrefix(s, "'") {
+		v, err := parseQuotedValue(s, '\'', false)
+		return v, err
+	}
+	// Unquoted value: stop before an unquoted comment start '#'
+	// when it's the first character or preceded by whitespace.
+	for i := 0; i < len(s); i++ {
+		if s[i] == '#' {
+			if i == 0 || isSpace(s[i-1]) {
```
kostko, line 101: Join these conditions.

cmd/rofl/dotenv.go
```diff
@@ -0,0 +1,190 @@
...
+// collectQuotedValue accumulates a possibly multi-line quoted value until the closing quote.
+// Returns the parsed value, the number of extra lines consumed (beyond the current one), or an error.
+func collectQuotedValue(valuePart string, lines []string, startIdx int) (string, int, error) {
+	delim := valuePart[0]
+	unescape := delim == '"'
+	combined := valuePart
+	consumed := 0
+
+	for {
+		v, e := parseQuotedValue(combined, delim, unescape)
```
kostko, line 123:
```suggestion
		v, err := parseQuotedValue(combined, delim, unescape)
```

kostko (approved):

PR #577, opened 2025-08-21
Fix artifact upgrade and bump rofl-containers

### LINKED ISSUE #576: `rofl upgrade` should not overwrite unchanged artifacts
Currently running `rofl upgrade` will result in some artifacts that are not part of the upgrade (eg. `builder` and `compose`) to be incorrectly overwritten.

kostko:
Closes #576

### DIFF
--- a/build/rofl/manifest.go
+++ b/build/rofl/manifest.go
@@ -500,10 +500,55 @@ type ArtifactsConfig struct {
 	Container ContainerArtifactsConfig `yaml:"container,omitempty" json:"container,omitempty"`
 }
 
+type artifactUpgrade struct {
+	existing *string
+	new      string
+}
+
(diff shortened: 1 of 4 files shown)


build/rofl/manifest.go
```diff
@@ -500,10 +500,55 @@ type ArtifactsConfig struct {
...
+	}
+	return changed
+}
+
+// UpgradeTo upgrades the artifacts to the latest version by updating any relevant fields.
+//
+// Returns true iff any artifacts have been updated.
+func (ac *ArtifactsConfig) UpgradeTo(latest *ArtifactsConfig) bool {
+	var changed bool
+	changed = upgradeArtifacts([]artifactUpgrade{
```
ptrus, line 528: `|| changed` is unneeded here.

ptrus (approved):

PR #587, opened 2025-08-28
fix: also resolve outputDir in extractArchive

njelich:
On macOS Docker /var is a symlink to /private/var, so we need to resolve the outputDir to prevent errors when checking the prefix matching.


njelich:
Update: The hang occurs because of the download of the rofl-dev image that is used in the WrapCommand utility taking a long time without a message to the user. It would maybe be useful to pipe the actual docker pull progres sso the user knows what is happening (this step could easily take more than 10 minutes with no additional output nor information on what is going on).

One idea is also to make a separate step to pull the images.

njelich:
@kostko We found the issue to be that the image being used as a base for the rofl-dev image is the very large and bloated `oasis-core-dev` [image](https://github.com/oasisprotocol/oasis-sdk/blob/76c1708c37d8d6be7f5d17ba795fdaf7aa039dae/docker/rofl-dev/Dockerfile#L1). Do we need anything from Oasis core image in the ROFl build process or could we replace it with something slimmer (and maybe alpine based)?

kostko:
And yeah it would be great if initial pull could show progress. Maybe environment init should pull the image and show progress?

njelich:
@kostko Please reapprove and merge.

tjanez:
@njelich , please never merge `master` branch into your own branch, this is bad git higiene.

Rather, rebase your commits on top of the `master` branch.

tjanez (approved):

PR #588, opened 2025-08-29
Show progress bar when uploading to OCI repository

### LINKED ISSUE #522: `rofl deploy`: Show a progress bar when uploading orc to OCI repo
.orc bundles are currently ~80MB big and on slower connections it may take some time to upload. Show a progress bar (similar to what `wget` has?) to show that the .orc is still uploading.

abukosek:
Closes #522.


cmd/common/progress/progress.go
```diff
@@ -1,12 +1,52 @@
...
 	"fmt"
 	"io"
+	"math"
 	"os"
+	"strings"
 	"time"
 )
 
+// PrintProgressBar prints a nice progress bar.
+func PrintProgressBar(f io.Writer, msg string, done uint64, total uint64, final bool) {
```
matevz, line 14: Could you use the percentage on the left, then the barline and on the right amount of data in MiB?

Something like wget:

image

Later we can also add the current speed and an eta.

matevz (approved):
One more suggestion, otherwise looks good!

PR #605, opened 2025-09-10
feat(cmd/rofl): Add support for machine permissions

kostko:


build/rofl/manifest.go
```diff
@@ -408,6 +408,9 @@ type Machine struct {
 	Offer string `yaml:"offer,omitempty" json:"offer,omitempty"`
 	// ID is the identifier of the machine to deploy into.
 	ID string `yaml:"id,omitempty" json:"id,omitempty"`
+
+	// Permissions is a map of permissions for the machine.
+	Permissions map[string][]string `yaml:"permissions,omitempty" json:"permissions,omitempty"`
```
matevz, line 413: Is simple whitelisting specific addresses enough? Could there be a permission where you'd open it to anyone? (and the other way around - is preventing anyone from doing an action done by simply not putting a permission to this map)

PR #610, opened 2025-09-16
feat(cmd/rofl): Add support for custom proxy domains

kostko:


cmd/rofl/build/validate.go
```diff
@@ -132,23 +160,40 @@ func validateComposeFile(composeFile string, manifest *buildRofl.Manifest) error
...
+
+			if port.Protocol != "tcp" {
+				continue
+			}
+			proxyMode := service.Annotations[fmt.Sprintf("net.oasis.proxy.ports.%s.mode", port.Published)]
+			if proxyMode == "ignore" {
+				continue
+			}
+			genericDomain := fmt.Sprintf("p%s", port.Published)
+			customDomain := service.Annotations[fmt.Sprintf("net.oasis.proxy.ports.%s.custom_domain", port.Published)]
```
matevz, line 189: is it wise to put the domain name inside `compose.yaml` instead of `rofl.yaml`? Suppose you have a testing and a production version of your app. Then you'd need two separate compose files in this case.

kostko: Hm, yeah I wanted to simplify the amount of configuration for the user.

It definitely needs to be set in `compose.yaml` because you can have a domain for each port and it makes sense to have that configuration in one place.

But what you said also makes sense. Maybe we should just have per-deployment artifacts?

kostko: No that would currently not work as the parser is not doing environment variable substitution when parsing the YAML compose file.

matevz (approved):

PR #613, opened 2025-09-17
cmd/rofl/deploy: Also show a "note" metadata when listing offers

matevz:
This PR shows a "note" metadata of the offer:
image

Hardcoding which metadata is shown (currently it's just "note") instead of showing everything is a good idea IMO to avoid potential attacks (e.g. "price" metadata).


cmd/rofl/deploy.go
```diff
@@ -413,6 +413,9 @@ func showProviderOffer(ctx context.Context, offer *roflmarket.Offer) {
 		offer.Resources.CPUCount,
 		float64(offer.Resources.Storage)/1024.,
 	)
+	if _, ok := offer.Metadata["note"]; ok {
```
kostko, line 416: I would use namespaces for metadata, eg. `net.oasis.note` instead of a global "note".

kostko (approved):

PR #629, opened 2025-10-16
cmd/wallet/create: set default name

csillag:
This makes it possible to execute `oasis wallet create` without any arguments, and get a wallet called `test_wallet`.

The goal of this change is to get our CLI tool more inline with expectations (hallucinations) of AI coding tools, in order to decrease friction for new developers.

### DIFF
--- a/cmd/wallet/create.go
+++ b/cmd/wallet/create.go
@@ -15,12 +15,15 @@ import (
 var accKind string
 
 var createCmd = &cobra.Command{
-	Use:   "create <name>",
+	Use:   "create [<name>]",
 	Short: "Create a new account",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.MaximumNArgs(1),
 	Run: func(_ *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


cmd/wallet/create.go
```diff
@@ -15,12 +15,15 @@ import (
...
 var createCmd = &cobra.Command{
-	Use:   "create <name>",
+	Use:   "create [<name>]",
 	Short: "Create a new account",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.MaximumNArgs(1),
 	Run: func(_ *cobra.Command, args []string) {
 		cfg := config.Global()
-		name := args[0]
+		name := "test_wallet"
```
kostko: Maybe it should instead be called "default" or similar?

matevz (approved):

matevz (commented):
Can you also please update the docs https://github.com/oasisprotocol/cli/blob/master/docs/wallet.md#create-an-account-create

PR #637, opened 2025-11-03
cmd/statesync: Add command displaying recommended trust root

### LINKED ISSUE #558: Add command displaying recommended light client's trust for the state sync
**Motivation**:
The process of finding correct light client trust needed for the consensus state sync is currently manual and tedious.

**Feature request**:
Add a new command to the oasis cli that displays appropriate light client trust.

**Context**:
1. Light client trust period must be lower than debonding period (336 epoch, ~2 weeks) to prevent long range attacks.
2. Checkpointing frequency is defined by the consensus parameters and currently happens 1/week.
    *  I.e. unless manually modified all nodes create checkpoints at the same heights.
3. Node does a state sync by first requesting available checkpoints from its peers. To prevent spam, it first validates advertised checkpoint and the corresponding root by comparing it to the light client root at a given height. 

As a consequence, one must configure _sufficiently old_ light client trust (height, hash and period), so that there will be corresponding light client header available for at least one checkpoint advertised by the p2p network. Finally, trust period must be be longer than the age of configured trusted header, else a checkpoint that happened recently would fall out of trust period interval.

**Proposed solution**:
Query consensus parameters (debonding period, block frequency, checkpoint interval). Calculate appropriate trusted period and trusted height. Finally, query block header for the trusted height (trusted hash). 

For the current set of parameters, I would expect this command to return trust period of ~12 days, and trusted height that is ~10 days old.

**Alternative solution**:
Less optimal, once https://github.com/oasisprotocol/oasis-core/issues/5823 is merged you will be able to query node's local checkpoint heights from the status.  In case node is not creating checkpoints we could add a separate method to the node that queries checkpoints using its p2p client. This seems quite a lot of work for smtg that can be done simpler.

martintomazic:
Closes #558

## Considerations:
https://github.com/oasisprotocol/cli/pull/637#discussion_r2488049630

## How to test
Run
```bash
./oasis state-sync trust  --network testnet|mainnet
```
end ensure you are able to do state sync with returned parameters.

### DIFF
--- a/cmd/network/trust.go
+++ b/cmd/network/trust.go
@@ -20,6 +20,11 @@ const (
 	// be able to fetch and verify light headers and finally detect and penalize a possible
 	// byzantine behavior. This value intentionally overestimates the actual time required.
 	maxVerificationTime = 24 * time.Hour
+
+	// maxCpCreationTime defines the upper bound within which remote node should produce
+	// checkpoint. This value intentionally overestimates the actual time required,
+	// assuming nodes with very bulky NodeDB, where checkpoint creation is slow.
+	maxCpCreationTime = 36 * time.Hour
 )
 
 var trustCmd = &cobra.Command{
@@ -94,11 +99,12 @@ func calcTrust(ctx context.Context, conn connection.Connection) (config.TrustCon
 	}
 	trustPeriod := calcTrustPeriod(debondingPeriod)
 
-	// Going back the whole checkpoint interval guarantees there should be at least
-	// one target checkpoint height inside this interval.
-	candidate, err := conn.Consensus().Core().GetBlock(ctx, latest.Height-cpInterval)
+	// Going back the whole checkpoint interval plus max checkpoint creation time guarantees there wil be at least
+	// one target checkpoint height from the trust height onwards, for which remote nodes already created a checkpoint.
+	candidateHeight := latest.Height - cpInterval - int64(maxCpCreationTime/blkTime)
+	candidate, err := conn.Consensus().Core().GetBlock(ctx, candidateHeight)
 	if err != nil {
-		return config.TrustConfig{}, fmt.Errorf("failed to get candidate trust (height: %d): %w", candidate.Height, err)
+		return config.TrustConfig{}, fmt.Errorf("failed to get candidate trust (height: %d): %w", candidateHeight, err)
 	}
 
 	// Sanity check: the trusted root must not be older than the sum of the trust


cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Sanity check: the trusted root must not be older than the sum of the trust
+	// period and the maximum verification time. If it is, the light client will
+	// not be able to safely use this block as a trusted root for the state sync.
+	if candBlk.Time.Add(trustPeriod).Add(maxVerificationTime).Before(time.Now()) {
+		return trust, fmt.Errorf("failed to verify candidate trust root")
+	}
+
+	trust.Period = trustPeriod
+	trust.Height = uint64(candHeight) // #nosec G115
```
martintomazic: NIT: I assume there is no clean way to avoid `#nosec G115`?

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+
+func calcAvgBlkTime(ctx context.Context, conn connection.Connection, latest *api.Block) (time.Duration, error) {
+	var deltaBlocks int64 = 1000
+	blk, err := conn.Consensus().Core().GetBlock(ctx, latest.Height-deltaBlocks)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get latest block: %w", err)
+	}
+
+	return latest.Time.Sub(blk.Time) / time.Duration(deltaBlocks), nil
+}
```
martintomazic: Alternative is to use `GetLastRetainedHeight`. Still both options assume block production is stable. In case the chain halts this may return wrong results... But then we have a bigger problem.


Alternative is to query the local checkpoint heights from the status. As already [discussed](https://github.com/oasisprotocol/cli/issues/558#issuecomment-3198495129) this is not realistic as public grcp nodes do not produce checkpoints, so for the majority of users this will be of little help. There is even more [complicated solution](https://github.com/oasisprotocol/oasis-core/pull/6265#discussion_r2238032326), of fetching checkpoint heights from the p2p that I would ideally like to avoid.

We could even use combination of both and fallback on the calculation only if not available from the connection provider status. Again I believe this complicates things so would prefer to keep it simple if we agree.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/oasisprotocol/oasis-core/go/consensus/cometbft/config"
+
+	"github.com/oasisprotocol/cli/cmd/common"
+	cliConfig "github.com/oasisprotocol/cli/config"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+	"github.com/spf13/cobra"
+)
+
+const (
+	// maxVerificationTime significantly overestimates the time needed for the
```
peternose: This comment could be better structured, e.g. first description what this is, and then that it overestimates ...

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	// should be significantly smaller than the total debonding period.
+	trustPeriodFraction = 3.0 / 4.0
+)
+
+var trustCmd = &cobra.Command{
+	Use:   "trust",
+	Short: "Show the recommended light client trust for consensus state sync",
+	Args:  cobra.NoArgs,
+	RunE: func(_ *cobra.Command, _ []string) error {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
peternose: Doesn't cobra have come kind of ctx which cancels if I cancel command?

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	},
+}
+
+// calculateTrustCfg calculates and verifies the recommended trust config.
+func calcTrustCfg(ctx context.Context, conn connection.Connection) (config.TrustConfig, error) {
+	var trust config.TrustConfig
+
+	latestBlk, err := conn.Consensus().Core().GetBlock(ctx, api.HeightLatest)
+	if err != nil {
+		return trust, fmt.Errorf("failed to get latest block: %w", err)
```
peternose:
```suggestion
		return nil, fmt.Errorf("failed to get latest block: %w", err)
```

I prefer `nil`, as I'm not sure if `trust` has zero value.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	latestHeight := latestBlk.Height
```
peternose: This var is not needed if you simplify naming, e.g. `latestBlk` to `latest`, `candBlk` to `candidate` or `trusted`, ...

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	debondingPeriod, err := calcDebondingPeriod(ctx, conn, latestHeight, blkTime)
+	if err != nil {
+		return trust, fmt.Errorf("failed to calculate debonding period (height: %d): %w", latestHeight, err)
+	}
+	trustPeriod := calcTrustPeriod(debondingPeriod)
+
+	// Going back the whole checkpoint interval guarantees there will be at least
+	// one checkpoint available.
```
peternose: Might not be true. It guaranties there was one checkpoint height. You cannot be sure if there are checkpoints on the network.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	candBlk, err := conn.Consensus().Core().GetBlock(ctx, candHeight)
+	if err != nil {
+		return trust, fmt.Errorf("failed to get candidate trust (height: %d): %w", candHeight, err)
+	}
+
+	// Sanity check: the trusted root must not be older than the sum of the trust
+	// period and the maximum verification time. If it is, the light client will
+	// not be able to safely use this block as a trusted root for the state sync.
+	if candBlk.Time.Add(trustPeriod).Add(maxVerificationTime).Before(time.Now()) {
+		return trust, fmt.Errorf("failed to verify candidate trust root")
```
peternose: Not the best error message.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+		return trust, fmt.Errorf("failed to calculate average block time: %w", err)
+	}
+
+	debondingPeriod, err := calcDebondingPeriod(ctx, conn, latestHeight, blkTime)
+	if err != nil {
+		return trust, fmt.Errorf("failed to calculate debonding period (height: %d): %w", latestHeight, err)
+	}
+	trustPeriod := calcTrustPeriod(debondingPeriod)
+
+	// Going back the whole checkpoint interval guarantees there will be at least
```
peternose: Can we figure out the exact block height? Using that would be better as the block is the closest to the checkpoint, meaning that the operator can use this trust root for longer period of time.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+	// one checkpoint available.
+	if cpInterval > latestHeight {
+		return trust, fmt.Errorf("checkpoint interval exceeds latest height")
+	}
+	candHeight := latestHeight - cpInterval
+	candBlk, err := conn.Consensus().Core().GetBlock(ctx, candHeight)
+	if err != nil {
+		return trust, fmt.Errorf("failed to get candidate trust (height: %d): %w", candHeight, err)
+	}
```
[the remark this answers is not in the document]
peternose: What would be the error if checkpoint is in progress and not available?
Should we add a parameter how many checkpoints you want to capture?

cmd/statesync/statesync.go
```diff
@@ -0,0 +1,16 @@
...
+package statesync
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// Cmd is the statesync sub-command.
+var Cmd = &cobra.Command{
+	Use:   "state-sync",
+	Short: "State sync utilities for the consensus layer",
```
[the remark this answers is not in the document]
martintomazic: By `consensus` you actually mean `network` or new `consensus` section?

cmd/statesync/trust.go
martintomazic: Still good practice would be to define expected timeout in which this command should finish?

cmd/statesync/trust.go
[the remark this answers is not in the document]
peternose: I would for now avoid `maxCpCreationTime`, and lets fix it once we encounter problems.

cmd/statesync/statesync.go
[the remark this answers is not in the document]
peternose: Sorry, I meant `network`.

cmd/statesync/trust.go
peternose: Yes, I was just thinking about replacing `context.Background()`.

cmd/statesync/trust.go
```diff
@@ -0,0 +1,157 @@
...
+func fetchCheckpointInterval(ctx context.Context, conn connection.Connection, height int64) (int64, error) {
+	params, err := conn.Consensus().Core().GetParameters(ctx, height)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get consensus parameters (height: %d): %w", height, err)
+	}
+	return int64(params.Parameters.StateCheckpointInterval), nil // #nosec G115
+}
+
+func calcAvgBlkTime(ctx context.Context, conn connection.Connection, latest *api.Block) (time.Duration, error) {
+	var deltaBlocks int64 = 1000
```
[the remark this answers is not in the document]
peternose: You can define it inline, e.g. `const pi = 3.14159 `.

cmd/statesync/trust.go
peternose: If formula changes, you can change it also here. It is like changing API. And we will notice when formula changes as the command won't work anymore for state sync.

cmd/statesync/trust.go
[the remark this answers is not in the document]
martintomazic: >I would for now avoid maxCpCreationTime, and lets fix it once we encounter problems.

Pointing this out because I have already experienced this in the past when setting up only a week old trust, thinking it should work but then when debugging `w.getCheckpointList(ctx)` inside `checkpoint_syn.go` I came to the conclusions above. Will check tmr again if things changed...

cmd/network/trust.go
```diff
@@ -0,0 +1,166 @@
...
+	cliConfig "github.com/oasisprotocol/cli/config"
+)
+
+const (
+	// maxVerificationTime defines the upper bound within which the light client should
+	// be able to fetch and verify light headers and finally detect and penalize a possible
+	// byzantine behavior. This value intentionally overestimates the actual time required.
+	maxVerificationTime = 24 * time.Hour
+
+	// maxCpCreationTime defines the upper bound within which remote node must produce
```
peternose:
```suggestion
	// maxCpCreationTime defines the upper bound within which remote node should produce
```

cmd/network/trust.go
```diff
@@ -0,0 +1,166 @@
...
+
+	debondingPeriod, err := calcDebondingPeriod(ctx, conn, latest.Height, blkTime)
+	if err != nil {
+		return config.TrustConfig{}, fmt.Errorf("failed to calculate debonding period (height: %d): %w", latest.Height, err)
+	}
+	trustPeriod := calcTrustPeriod(debondingPeriod)
+
+	// Going back the whole checkpoint interval plus max checkpoint creation time guarantees there wil be at least
+	// one target checkpoint height from the trust height onwards, for which remote nodes already created a checkpoint.
+	candidate, err := conn.Consensus().Core().GetBlock(ctx, latest.Height-cpInterval-int64(maxCpCreationTime/blkTime))
```
peternose: Doing 3 things in a very long line, should be shortened.

PR #647, opened 2025-11-15
Fix oasis rofl build for manfiests using a builder env

ptrus:
Found out while doing: https://github.com/oasisprotocol/oasis-sdk/pull/2402

Also confirmed this still works for ROFL apps that don't use the build env.

### DIFF
--- a/cmd/rofl/build/artifacts.go
+++ b/cmd/rofl/build/artifacts.go
@@ -359,12 +359,18 @@ func createSquashFs(buildEnv env.ExecEnv, fn, dir string) (int64, error) {
 	tarHash := hex.EncodeToString(tarHasher.Sum(nil))
 	fmt.Printf("TAR archive SHA256: %s\n", tarHash)
 
+	// Convert paths for container environment if needed.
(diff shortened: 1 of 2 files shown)


cmd/rofl/build/artifacts.go
```diff
@@ -359,36 +359,39 @@ func createSquashFs(buildEnv env.ExecEnv, fn, dir string) (int64, error) {
...
+		return 0, fmt.Errorf("failed to convert output path: %w", err)
+	}
+	tarPathInEnv, err := buildEnv.PathToEnv(tarPath)
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert tar path: %w", err)
+	}
+
 	// Convert tar to squashfs using sqfstar under fakeroot.
+	// We use shell redirection instead of cmd.Stdin because stdin doesn't get piped through
+	// Docker when running in a container environment.
```
kostko: Hm isn't that strange, maybe it needs a flag to pipe?

ptrus: You were right, we just needed `-i` for docker. Updated, since that's much nicer fix. Thanks!

PR #654, opened 2025-11-26
docs/network: Fix broken link

martintomazic:
Fixes https://github.com/oasisprotocol/docs/actions/runs/19626086942/job/56195086882?pr=1574.

(I originally copied the pattern from the sub-chapters, but in case of the heading no anchor will be created).


matevz:
No need to reference the anchor of the top-level heading yeah.

Issue #652, opened 2025-11-24
`oasis rofl build` without `builder` should just fail on macOS


matevz:
This was suggested by one of our hackathoners:

Since `oasis rofl build` cannot work natively on macOS (and windows?) if no `artifacts.builder` is defined, it should simply crash saying sth like

```
Error: Building ROFL natively on {platform.os}_{platform.architecture} is not supported yet. Invoke the command inside a container:
docker/podman run --platform linux/amd64 --volume .:/src -it ghcr.io/oasisprotocol/rofl-dev:main oasis rofl build ... and any flags
```

Currently it goes on complaining that some library is missing etc.

kostko:
Why not say to configure the builder and/or make builder the default?

ptrus:
Depends on which option we prefer (on macOS):

1) `oasis rofl build` works, but is not reproducible with Linux (even if using artefacts.builder) #657 implements this
2) `oasis rofl build` only works in cases where it's reproducible (so Linux, or within Docker on other platforms). It fails in other cases, suggesting the appropriate command
  - This is a simple change
3) `oasis rofl build` works and produces reproducible images
  - I think the only way to achieve this realistically is that `oasis rofl build` self-executes itself within Docker/Podman when on non-Linux. Basically the automated version of 2). But is a more complex/bigger change

PR #657, opened 2025-11-27
Configure builder by default

ptrus:
Fixes: https://github.com/oasisprotocol/cli/issues/652

This PR configures the ROFL builder container image by default for new and upgraded ROFL apps. Also provides clearer error messages when the container runtime is missing or native builds aren't supported.

Everyone should now just run `oasis rofl build` (regardless of platform).

Note: This changes the built artifacts again, but builds should now be fully reproducible across environments when using the Docker builder.

Changes:
- Builder container image is now configured by default for new and upgraded apps
- Stream tar directly to sqfstar instead of extracting to disk
- Split builder images: minimal rofl-container-builder for container apps, full rofl-dev for raw Rust apps

Tested this (and got reproducable builts on):
- ubuntu with `oasis rofl build --no-container` (no builder)
- ubuntu with `oasis rofl build` (uses builder)
- macos with builder `oasis rofl build` (uses builder)
 
TODO: 
- [x] Consider adding a CLI version field to the manifest (set on init/upgrade) to track which CLI version created/modified it.
- [x] update [quickstart docs](https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/quickstart.mdx#build) to not recommend `docker run --platform linux/amd64 --volume .:/src -it ghcr.io/oasisprotocol/rofl-dev:main oasis rofl build` anymore, once this is released
  - done in https://github.com/oasisprotocol/oasis-sdk/pull/2411

### DIFF
--- a/build/rofl/artifacts.go
+++ b/build/rofl/artifacts.go
@@ -1,7 +1,12 @@
 package rofl
 
-// LatestBuilderImage is the latest builder container image to use when building ROFL apps.
-const LatestBuilderImage = "ghcr.io/oasisprotocol/rofl-dev:v0.5.0@sha256:31573686552abeb0edebc450f6872831f0006a6cf38220cef7e0789d4376c2c1"
+// Builder images for different app kinds.
+const (
+	// LatestBuilderImage is the full builder with Rust toolchain for raw apps.
+	LatestBuilderImage = "ghcr.io/oasisprotocol/rofl-dev:v0.5.0@sha256:31573686552abeb0edebc450f6872831f0006a6cf38220cef7e0789d4376c2c1"
+	// LatestContainerBuilderImage is the minimal builder for container apps.
+	LatestContainerBuilderImage = "ghcr.io/oasisprotocol/rofl-container-builder:0.0.1@sha256:913ef97ab07dde31f08ce873f825bf3d4f32ad4102ff5797d7c3050c121c4dce"
+)
 
 // LatestBasicArtifacts are the latest TDX ROFL basic app artifacts.
 var LatestBasicArtifacts = ArtifactsConfig{
--- a/cmd/rofl/build/artifacts.go
+++ b/cmd/rofl/build/artifacts.go
@@ -8,13 +8,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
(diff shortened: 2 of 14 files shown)


build/rofl/artifacts.go
```diff
@@ -1,5 +1,8 @@
 package rofl
 
+// LatestBuilderImage is the latest builder container image to use when building ROFL apps.
+const LatestBuilderImage = "ghcr.io/oasisprotocol/rofl-dev:v0.5.0@sha256:31573686552abeb0edebc450f6872831f0006a6cf38220cef7e0789d4376c2c1"
```
ptrus: Actually, this is a bit awkward, because `rofl-dev` contains CLI. So we need to release a new CLI version to bump it in the rofl-dev, but then we need to update the ref in CLI. Hm but do we even need CLI there?

matevz: `rofl-dev` is also for container-everything cases where people don't want to install anything on their PC: https://docs.oasis.io/build/rofl/workflow/prerequisites#conservative-containers-everywhere

ptrus: Yes, we should use a different builder and build it in this repository.

cmd/rofl/build/artifacts.go
```diff
@@ -576,8 +615,27 @@ func convertToQcow2(buildEnv env.ExecEnv, fn string) error {
 		return fmt.Errorf("%w\n%s", err, out.String())
 	}
 
+	// Resize to ensure virtual size matches original (prevents truncation).
```
kostko: Was this also a problem before?

kostko (approved):
Nice work!

cmd/rofl/build/artifacts.go
ptrus: I figured out what the issue was. It wasn't related to the environment. The problem was that the padding was done on the host (not in the builder container), and over the bind mount the container saw only the pre‑pad size. Doing either `veritysetup verify` or resizing the image fixed the issue (probably just due to touching it?). So I moved the padding to be run in the builder env and it fixes the issue. I have kept the `veritysetup verify` step, since it's useful as a sanity check, but have removed the resize since it's not needed.

PR #695, opened 2026-04-17
cmd/network/show: Print entity name

### LINKED ISSUE #633: Add validator names to `oasis network show committees`

matevz:
Fixes https://github.com/oasisprotocol/cli/issues/633

Vibe coded with qwen3.6:35b.

### DIFF
--- a/cmd/network/show.go
+++ b/cmd/network/show.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/oasisprotocol/cli/cmd/common"
 	cliConfig "github.com/oasisprotocol/cli/config"
+	"github.com/oasisprotocol/cli/metadata"
 	"github.com/oasisprotocol/cli/table"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/network/show.go
```diff
@@ -306,12 +307,40 @@ var showCmd = &cobra.Command{
 				})
 				cobra.CheckErr(err)
 
+				// Build a lookup map from entity ID to entity name.
+				fromRegistry, err := metadata.EntitiesFromRegistry(ctx)
+				if err != nil {
+					common.Warnf("Warning: failed to query metadata registry: %v\n", err)
+				}
+				fromOasisscan, err := metadata.EntitiesFromOasisscan(ctx)
```
peternose, line 315: Why do you need data also from Oasis scan?

cmd/network/show.go
```diff
@@ -306,12 +307,40 @@ var showCmd = &cobra.Command{
...
+				}
+				entityNameByID := make(map[string]string)
+				for _, src := range []struct {
+					m      *map[types.Address]*metadata.Entity
+					suffix string
+				}{
+					{&fromRegistry, ""},
+					{&fromOasisscan, " (from oasisscan)"},
+				} {
+					if src.m == nil || *src.m == nil {
```
peternose, line 327: Hm ... 🤔 

```go
				// Build a lookup map from entity ID to entity name.
				fromRegistry, err := metadata.EntitiesFromRegistry(ctx)
				if err != nil {
					common.Warnf("Warning: failed to query metadata registry: %v\n", err)
				}
				fromOasisscan, err := metadata.EntitiesFromOasisscan(ctx)
				if err != nil {
					common.Warnf("Warning: failed to query oasisscan: %v\n", err)
				}
				entityNameByID := make(map[string]string)
				for _, src := range []struct {
					m      map[types.Address]*metadata.Entity
					suffix string
				}{
					{fromRegistry, ""},
					{fromOasisscan, " (oasisscan)"},
				} {
					for _, ent := range src.m {
						id := ent.ID.String()
						if _, ok := entityNameByID[id]; ok {
							continue
						}
						entityNameByID[id] = ent.Name + src.suffix
					}
				}

```

cmd/network/show.go
```diff
@@ -306,12 +307,40 @@ var showCmd = &cobra.Command{
...
+					}
+				}
+
 				for _, runtime := range runtimes {
 					if runtime.Kind != registry.KindCompute {
 						continue
 					}
 					table := table.New()
-					table.Header("Entity ID", "Node ID", "Role")
+					table.Header("Entity Name", "Entity ID", "Node ID", "Role")
```
peternose, line 343: Since the entity name is not to be trusted and can be empty, I would not put it in the first column.

cmd/network/show.go
```diff
@@ -337,7 +366,13 @@ var showCmd = &cobra.Command{
 						node, err := registryConn.GetNode(ctx, nodeQuery)
 						cobra.CheckErr(err)
 
+						name := entityNameByID[node.EntityID.String()]
+						if name == "" {
+							name = "unknown"
```
peternose, line 371: I would just use `""`, as one could have an entity with name `unknown`.

cmd/network/show.go
```diff
@@ -306,12 +307,40 @@ var showCmd = &cobra.Command{
...
+					{&fromOasisscan, " (from oasisscan)"},
+				} {
+					if src.m == nil || *src.m == nil {
+						continue
+					}
+					for _, ent := range *src.m {
+						if _, exists := entityNameByID[ent.ID.String()]; exists {
+							continue
+						}
+						entityNameByID[ent.ID.String()] = ent.Name + src.suffix
```
peternose, line 334: If the name is empty, would this produce name `" (from oasiscan)"`?

PR #698, opened 2026-05-20
feat: add support for per-deployment ROFL artifacts

### LINKED ISSUE #697: Add support for per-deployment `artifacts`
ROFL secrets are (and must be) stored per-deployment.

In case you want to deliberately expose certain environment variables, `compose.yaml` is the place to put those where everyone can check them. But currently, you can only define a single global compose file per ROFL manifest file.

Enable the `artifacts` section to also appear inside each deployment section. In this case Oasis CLI should combine it with the globally defined `artifacts`. This will enable developer to define deployment-specific artifacts including the compose file and public environment variables.

uniyalabhishek:
closes https://github.com/oasisprotocol/cli/issues/697

Adds support for deployment-level artifacts in ROFL manifests, resolving effective artifacts as defaults -> global manifest artifacts -> selected deployment overrides. 

This lets each deployment use its own compose file, including public deployment-specific environment config.

Tested a real Testnet ROFL build/update/deploy using a deployment-specific compose file, the deployed app served the expected deployment-specific port/content through the ROFL proxy.

### DIFF
--- a/docs/rofl.md
+++ b/docs/rofl.md
@@ -103,34 +103,6 @@ chapter for details.
 [ROFL Prerequisites]: https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/workflow/prerequisites.md
 [npa]: ./account.md#npa
 
-### Deployment-specific artifacts {#deployment-artifacts}
-
-Artifacts can be configured globally or per deployment. Deployment artifacts are
-overlaid field by field on top of global artifacts, so a deployment can override
-only the compose file while reusing the same firmware, kernel, stage 2 image and
-container runtime.
-
-This is useful when public deployment-specific environment variables belong in
-the compose file while encrypted secrets remain in the selected deployment
-section:
-
-```yaml
-artifacts:
-  container:
-    compose: compose.yaml
-
-deployments:
-  testnet:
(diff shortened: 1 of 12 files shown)


docs/rofl.md
```diff
@@ -103,6 +103,34 @@ chapter for details.
 [ROFL Prerequisites]: https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/workflow/prerequisites.md
 [npa]: ./account.md#npa
 
+### Deployment-specific artifacts {#deployment-artifacts}
```
matevz: This should go under https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/features/manifest.md?plain=1#L83 for global `artifacts` and then again inside https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/features/manifest.md?plain=1#L89 mentioning `artifacts` can also be used per-deployment.

matevz (approved):

docs/rofl.md
uniyalabhishek: moved docs in sdk https://github.com/oasisprotocol/oasis-sdk/pull/2464

PR #699, opened 2026-05-28
cmd/rofl: Add set-trust-root subcommand

### LINKED ISSUE #567: Add subcommand to refresh ROFL app trust root
Currently the only way to refresh the trust root is to do so manually by editing the `rofl.yaml` file. We should add a simple subcommand, e.g. `oasis rofl refresh-trust-root` which fetches the latest trust root and updates the `rofl.yaml` app manifest.

anzoman:
Adds `oasis rofl refresh-trust-root` to fetch the latest consensus trust root and update the `rofl.yaml` manifest automatically.

Closes #567.


cmd/rofl/refresh_trust_root.go
```diff
@@ -0,0 +1,54 @@
...
+
+	buildRofl "github.com/oasisprotocol/cli/build/rofl"
+	"github.com/oasisprotocol/cli/cmd/common"
+	roflCommon "github.com/oasisprotocol/cli/cmd/rofl/common"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+)
+
+var refreshTrustRootCmd = &cobra.Command{
+	Use:   "refresh-trust-root",
```
matevz: For consistency with `oasis network set-chain-context` I would rather call this `set-trust-root` with block height and block hash as parameters. If you don't pass the parameters, then the latest height and hash is taken (your current implementation).

matevz:
While you're at it, can you also fix the `oasis rofl trust-root` output?

```
$ oasis rofl trust-root
TrustRoot {
    height: 32299682,
    hash: "a1f25ff6febef34216b08d31122bafc1fd24b0d2445a0f6b8da9b5f62624e89a".into(),
    runtime_id: "000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c".into(),
    chain_context: "0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76".to_string(),
}
```

Should be normal json.

cmd/rofl/refresh_trust_root.go
anzoman: Makes sense, updated as suggested.

cmd/rofl/set_trust_root.go
```diff
@@ -0,0 +1,82 @@
...
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+
+	buildRofl "github.com/oasisprotocol/cli/build/rofl"
+	"github.com/oasisprotocol/cli/cmd/common"
+	roflCommon "github.com/oasisprotocol/cli/cmd/rofl/common"
+)
+
+var setTrustRootCmd = &cobra.Command{
+	Use:   "set-trust-root [height] [hash]",
```
matevz:
```suggestion
	Use:   "set-trust-root [<height>] [<hash>]",
```

cmd/rofl/set_trust_root.go
```diff
@@ -0,0 +1,82 @@
...
+	Short: "Set the trust root in the rofl.yaml manifest",
```
kostko:
```suggestion
	Short: "Set the trust root in the ROFL app manifest",
```

cmd/rofl/set_trust_root.go
```diff
@@ -0,0 +1,82 @@
...
+	Long: "Set the consensus block height and hash of the trust root in the rofl.yaml manifest.\n\n" +
```
kostko:
```suggestion
	Long: "Set the consensus block height and hash of the trust root in the ROFL app manifest.\n\n" +
```

cmd/rofl/set_trust_root.go
anzoman: Updated.

cmd/rofl/set_trust_root.go
anzoman: Refactored.

cmd/rofl/set_trust_root.go
anzoman: Updated.

cmd/rofl/set_trust_root.go
```diff
@@ -0,0 +1,82 @@
...
+			deployment.TrustRoot = &buildRofl.TrustRootConfig{}
+		}
+		deployment.TrustRoot.Height = height
+		deployment.TrustRoot.Hash = hash
+
+		if err := manifest.Save(); err != nil {
+			cobra.CheckErr(fmt.Errorf("failed to update manifest: %w", err))
+		}
+
+		fmt.Printf("Updated '%s' with trust root: height=%d, hash=%s\n", manifest.SourceFileName(), height, hash)
```
matevz: I'd use the PrettyJSONMarshal we already use in the rest of the code for human-readable outputs. Then just pass the whole `deployment.TrustRoot` to be marshalled. Something like:

```suggestion
		out, err := json.PrettyJSONMarshal(deployment.TrustRoot)
		cobra.CheckErr(err)
		fmt.Printf("Updated '%s' with trust root:\n%s\n", manifest.SourceFileName(), string(out))
```

anzoman: Good catch. I have used `common.PrettyJSONMarshal` now.

matevz (approved):
Works as expected, thanks.

cmd/rofl/set_trust_root.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/connection"
+
+	buildRofl "github.com/oasisprotocol/cli/build/rofl"
+	"github.com/oasisprotocol/cli/cmd/common"
+	roflCommon "github.com/oasisprotocol/cli/cmd/rofl/common"
+)
+
+var setTrustRootCmd = &cobra.Command{
+	Use:   "set-trust-root [<height>] [<hash>]",
+	Short: "Set the trust root in the ROFL app manifest",
```
peternose, line 20: Command `oasis network trust` displays a warning `WARNING: Cannot be trusted unless the CLI is connected to the RPC endpoint you control.` when fetching trust root from the network. Should we also add one here?

anzoman: @peternose, good point, but that message is also not displayed when the ROFL app manifest is created along with the current trust root. Also, I think that warning should be more general and stated in our docs that the user using the CLI is aware that it is, by default, connected to the Oasis-managed RPC endpoints.

PR #705, opened 2026-07-05
Rofl public vars

martintomazic:
Part of https://github.com/oasisprotocol/oasis-sdk/issues/2248.

Used public var instead of env to make sure people don't accidentally set secrets here.

Tested it with https://github.com/oasisprotocol/Honoroll/pull/440.

### DIFF
--- a/build/rofl/artifacts.go
+++ b/build/rofl/artifacts.go
@@ -21,6 +21,6 @@ var LatestContainerArtifacts = ArtifactsConfig{
 	Kernel:   "https://github.com/oasisprotocol/oasis-boot/releases/download/v0.6.2/stage1.bin#e5d4d654ca1fa2c388bf64b23fc6e67815893fc7cb8b7cfee253d87963f54973",
 	Stage2:   "https://github.com/oasisprotocol/oasis-boot/releases/download/v0.6.2/stage2-podman.tar.bz2#b2ea2a0ca769b6b2d64e3f0c577ee9c08f0bb81a6e33ed5b15b2a7e50ef9a09f",
 	Container: ContainerArtifactsConfig{
-		Runtime: "https://github.com/oasisprotocol/oasis-sdk/releases/download/rofl-containers%2Fv0.8.6/rofl-containers#5aa26a3c5a7e1d284e217959d89b7a620ea7b7fc5079909e25042124ffb384e8",
+		Runtime: "https://github.com/oasisprotocol/oasis-sdk/releases/download/rofl-containers%2Fv0.9.0/rofl-containers#e2e074d03ab2fbacaacb01e6d63ce093fde04e4cc620dd8804e8afbb20390525",
 	},
 }
--- a/cmd/rofl/public_var.go
+++ b/cmd/rofl/public_var.go
@@ -1,0 +1,308 @@
+package rofl
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/spf13/cobra"
+
+	buildDotenv "github.com/oasisprotocol/cli/build/dotenv"
+	buildRofl "github.com/oasisprotocol/cli/build/rofl"
+	"github.com/oasisprotocol/cli/cmd/common"
+	roflCommon "github.com/oasisprotocol/cli/cmd/rofl/common"
+)
+
+const publicVarMetadataPrefix = "env."
+
+var (
+	publicVarCmd = &cobra.Command{
+		Use:   "public-var",
+		Short: "Public variable management commands",
+		Long: "Public variable management commands.\n\n" +
+			"Values are stored unencrypted in ROFL app metadata and exposed to containers as\n" +
+			"environment variables. Use `oasis rofl secret` for confidential values.",
+	}
+
+	publicVarSetCmd = &cobra.Command{
+		Use:   "set <name> <file>|-",
+		Short: "Set a public variable in the manifest, reading the value from file or stdin",
+		Args:  cobra.ExactArgs(2),
+		Run: func(_ *cobra.Command, args []string) {
+			publicVarName := args[0]
+			publicVarFn := args[1]
+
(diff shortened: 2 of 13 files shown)


cmd/rofl/public_var_test.go
```diff
@@ -0,0 +1,148 @@
...
+		testAPIURLValue   = "https://example.com?a=b#c"
+		testConfigVar     = "CONFIG"
+		testConfigValue   = `{"mode":"prod"}`
+	)
+
+	prevConfig := *cliConfig.Global()
+	cliConfig.ResetDefaults()
+	t.Cleanup(func() {
+		*cliConfig.Global() = prevConfig
+	})
```
martintomazic, line 60: NIT: existing global config makes things hard to test cleanly :((

build/rofl/artifacts.go
```diff
@@ -21,6 +21,7 @@ var LatestContainerArtifacts = ArtifactsConfig{
 	Kernel:   "https://github.com/oasisprotocol/oasis-boot/releases/download/v0.6.2/stage1.bin#e5d4d654ca1fa2c388bf64b23fc6e67815893fc7cb8b7cfee253d87963f54973",
 	Stage2:   "https://github.com/oasisprotocol/oasis-boot/releases/download/v0.6.2/stage2-podman.tar.bz2#b2ea2a0ca769b6b2d64e3f0c577ee9c08f0bb81a6e33ed5b15b2a7e50ef9a09f",
 	Container: ContainerArtifactsConfig{
+		// TODO release new rofl-container
```
martintomazic: Let's cut a new release and then I bump it here, followed by a new cli release.

martintomazic (commented):
Followed the style and organisation 1:1 with `oasis rofl secrets`.

kostko (approved):

docs/rofl.md
```diff
@@ -186,6 +186,53 @@ Run `rofl secret rm <secret-name>` to remove the secret from your manifest file.
 
 ![code shell](../examples/rofl/secret-rm.in.static)
 
+## Public variables management {#public-var}
+
+Public variables are stored unencrypted in ROFL app metadata and exposed to
```
matevz:
```suggestion
Public variables are stored unencrypted in ROFL app metadata and on-chain. They are exposed to
```

docs/rofl.md
```diff
@@ -186,6 +186,53 @@ Run `rofl secret rm <secret-name>` to remove the secret from your manifest file.
...
+containers as environment variables. Use `rofl secret` for confidential values.
```
matevz:
```suggestion
containers as environment variables. Use [`rofl secret`] to store confidential values.

[`rofl secret`]: #secret
```

matevz (commented):
If you define a public var with the same name as the existing secret, Oasis CLI should fail.

martintomazic:
>If you define a public var with the same name as the existing secret, Oasis CLI should fail.

Private variable name (if `--public-name` is set) is available inside `rofl.yaml` so the fix should work in all cases...? Writing tests for the new validation would require a thorough refactor of the existing code, so I only tested manually for now.

Note there is similar issue with secrets definitions overlapping between each other:
> the secret name is capitalized and spaces are replaced with underscores, so a secret called my secret will be available as MY_SECRET.

meaning cli should probably not allow setting `my secret` and `MY_SECRET` at the same time as a rofl secret, but I would ignore this as 1. existing and 2. app admin should be responsible.

matevz (commented):
Please also mention special `env.` metadata fields behavior in the ROFL manifest specs https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/features/manifest.md

cmd/rofl/public_var.go
```diff
@@ -0,0 +1,308 @@
...
+			}
+
+			fmt.Printf("Run `oasis rofl update` to update your ROFL app's on-chain configuration.\n")
+		},
+	}
+
+	// publicVarImportCmd bulk-imports public variables from a .env file (key=value with # comments).
+	// Supports '-' to read from stdin. Existing public variables are replaced only with --force.
+	publicVarImportCmd = &cobra.Command{
+		Use:   "import <dot-env-file>|-",
```
matevz, line 91: It would be useful to have another `rofl public-var set-env-file <dot-env-file>` command which references the given .env filename with the current deployment.

Then, ROFL manifest file is updated with something like

```yaml
deployments:
  testnet:
    metadata_env_file: .env.testnet
```

And when you run `oasis rofl update`, the content of this env file is read on-the-fly and populates `env.` fields in `metadata`. This way we never get inconsistency between .env files and the manifest.

martintomazic: I assume rofl github actions will have access to env files as they run in the repo checkout or? 

My concerns are:
1. What happens if you have variables defined both as part of the manifest metadata and inside manifest referenced env file? What happens if they overlap? Do we want precedence complexity?
2. Currently manifest acts as a source of truth for other commands, so every time `manifest.GetMetadata()` is called we will have to parse it add hoc into the same structure that will differ in memory to what is persisted? What happens when we later write the manifest back? Or we introduce overlay concept, that other commands need to learn about (complexity).
3. Now before running oasis rofl update you will have two sources of truth that you need to mentally merge.
4. Not all metadata is meant to act as public config to docker containers (only .env prefix).

Locally forgetting to run  `oasis rofl public-var import` is same as forgetting to write `oasis rofl update`? 

For the CI I would prefer to run a simple validation step, e.g. running `oasis rofl public-var import .env.testnet --force` should produce an empty diff for the CI to pass. This way you have sync guarantee without adding further complexity to the cli/manifest parsing?

martintomazic: Even better inside make/just file we add a command for sanity/sync validation, which can then be called locally (if deploying from local machine) or inside CI.

peternose:
> Please also mention special `env.` metadata fields behavior in the ROFL manifest specs https://github.com/oasisprotocol/oasis-sdk/blob/main/docs/rofl/features/manifest.md

You could also gather these variables under the same key, e.g. `metadata["env"] = "city=London,country=England"`.

matevz (approved):

PR #708, opened 2026-07-13
cmd/tx: Implement Safe transaction batch output format

### LINKED ISSUE #700: Safe integration
For multisig transactions [Safe](https://safe.oasis.io/) is probably the most widely used service today. It also runs on Sapphire Testnet and Mainnet. However, it only supports EVM transactions. ROFL and other Oasis-specific transactions need to be wrapped as subcalls inside ordinary EVM transaction if you want to sign them with Safe which introduces some friction.

Add support for Safe in Oasis CLI so that instead of broadcasting a transaction, it is wrapped and submitted to Safe for signers to sign and submit. This would standardize and simplify ROFL update process in CI for example:
- https://github.com/talos-agent/talos/blob/main/.github/workflows/release-rofl.yml
- https://github.com/talos-agent/talos/blob/main/scripts/propose_transactions/main.mjs

### Option 1: Safe account + API key

Apart from `File` and `Ledger` introduce `Safe` account type. Then store the Safe API key as a private key and a public multisig vault address as an address. Then when using `--account my_safe_account`, the CLI would send it to the Safe API instead of broadcacsting it to the chain.

### Option 2: Just export Safe formatted JSON file

Add `--format evm` used in combination with `--unsigned` and `-o my-rofl-update.hex`. Then upload this file to Safe manually.

matevz:
Implements option 2 of https://github.com/oasisprotocol/cli/issues/700

An example workflow for updating ROFL parameters on a ROFL that is governed by a multisig account on Safe:
1. `echo "abc" | oasis rofl public-var set GAS_PRICE -`
2. `oasis rofl update --format safe --unsigned -o update-rofl-gas-price.json`
3. Upload the file via [Safe Transaction Builder](https://safe.oasis.io/apps/open?safe=sapphire-testnet:0x51C96cb5Cd239B2913eA7bdde85A5EE9031baF1B&appUrl=https%3A%2F%2Fapps.safe.protofire.io%2Ftx-builder)
4. Sign the transaction with Safe (and wait for others to Sign)
5. Execute the transaction with Safe

### DIFF
--- a/cmd/common/safe.go
+++ b/cmd/common/safe.go
@@ -1,0 +1,134 @@
+package common
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	ethAbi "github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/config"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+// subcallAddress is the address of the "Subcall" precompile that dispatches
+// ABI-encoded (method, CBOR body) calldata as an SDK runtime call. See
+// https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/subcall.rs.
+const subcallAddress = "0x0100000000000000000000000000000000000103"
+
+// safeTxBuilderTransaction is a single call within a Safe Transaction Builder batch.
+type safeTxBuilderTransaction struct {
+	To                   string      `json:"to"`
+	Value                string      `json:"value"`
+	Data                 string      `json:"data"`
+	ContractMethod       interface{} `json:"contractMethod"`
+	ContractInputsValues interface{} `json:"contractInputsValues"`
+}
+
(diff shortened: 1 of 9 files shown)


cmd/common/safe.go
```diff
@@ -0,0 +1,132 @@
...
+	if !ok {
+		return nil, fmt.Errorf("--format safe is only supported for unsigned ParaTime transactions")
+	}
+	if network == nil {
+		return nil, fmt.Errorf("--format safe requires a network to be configured")
+	}
+	var chainID uint64 = 23294 // Use Safe on Sapphire Mainnet by default.
+	if network.ChainContext != config.DefaultNetworks.All["mainnet"].ChainContext {
+		chainID = 23295 // Sapphire Testnet
+	}
```
kostko: Should this do more validation of the paratime?

cmd/rofl/common/manifest.go
```diff
@@ -105,8 +105,8 @@ func MaybeLoadManifestAndSetNPA(cfg *cliConfig.Config, npa *common.NPASelection,
 	}
 	switch {
 	case d.Admin == "":
-		// No admin in manifest, leave unchanged.
-	case npa.AccountSetExplicitly:
+		// No admin in manifest, leave unchanged; ignore for unsigned transactions entirely.
+	case npa.AccountSetExplicitly, common.TxUnsigned:
```
matevz, line 109: This will be pretty common for multisig ROFLs where no one will have a local admin account available. But creating and exporting ROFL transactions into Safe format must still work.

kostko (approved):

PR #710, opened 2026-08-05
cmd/rofl: Support verifying versioned ROFL enclave IDs

matevz:
I'm not sure if this behavior of  `oasis rofl build --verify` is intentional, but the `--verify` flag only works, if the matched enclave ID does not have a version set.

Some projects may prefer to version all enclave IDs and not only the historical ones.


kostko (commented):
Shouldn't it compare the version with the latest configured version under the top-level `version` field instead of just using the highest version?

cmd/rofl/build/build.go
```diff
@@ -264,13 +264,23 @@ var (
 					}
 				}
 
-				if !maps.Equal(buildEnclaves, latestManifestEnclaves) {
-					fmt.Println("Built enclave identities DIFFER from latest manifest enclave identities!")
-					showIdentityDiff(buildEnclaves, latestManifestEnclaves, "Built", "Manifest")
+				// Any enclave identity present in the manifest (latest or versioned) is an
+				// acceptable match for the build.
```
kostko: It seems strange that one would verify against any version? Usually one would have the manifest be one for the version indicated in the `version` field and historic versions would be available from source control.

matevz: I enabled verification for both now - either the untagged one or the one defined in the ROFL version. The flow is not that clean anymore though. Maybe the previous behavior was better assuming that the untagged version is simply the one defined in the global version.

kostko: I think it makes sense that both mean the same? And maybe also versions higher than the `version` in the manifest should trigger some sort of warning?

PR #713, opened 2026-08-26
cmd/rofl/machine: Fix domain verification token derivation

anzoman:
### Problem

`oasis rofl machine show` prints the TXT record users must add to verify a custom proxy domain. Since #696 the token was derived from the **provider's scheduler app ID** instead of the **deployed app ID**:

```go
// fetched with out.Provider.SchedulerApp, to read the proxy domain metadata
out.Replica, _ = conn.Runtime(...).ROFL.AppInstance(ctx, ..., out.Provider.SchedulerApp, pk)
...
prettyPrintMachinePorts(mCfg.ExtraCfg, out.Replica.App, out.Machine, proxyDomain)
```

The scheduler derives it from `deployment.app_id`, so the record we tell users to add can never verify. The failure is opaque: the scheduler logs `TXT record not found` (a token *mismatch*, not a missing record), never installs the SNI mapping, and the app's ACME TLS-ALPN-01 challenge fails with `unexpected order status: Invalid` until Let's Encrypt's authorization-failure limit kicks in.

### Why it surfaced only now

#696 swapped in `out.Replica.App` to avoid a nil dereference when the machine has no deployment. That produced wrong tokens, but nothing failed visibly, because the scheduler's own derivation finalized TupleHash into a zero-length buffer and returned `""` for every input — which reduced verification to `contains("oasis-rofl-verification=")`, a prefix match that accepted any such record. `rofl-scheduler` v0.9.0 is the first release with the fixed derivation, so it is the first release that actually compares tokens.

### Fix

Derive from `insDsc.Deployment.AppID` again, keeping the nil handling #696 was after: when nothing is deployed we now print a short notice instead of a token.

### Impact

Anyone who added a TXT record based on output from a CLI containing #696 has an incorrect record and must re-run `oasis rofl machine show` and update DNS. Domains that appeared to verify under scheduler ≤0.8.0 were never genuinely verified.

### Testing

`gofmt`, `go vet`, `go test ./...` clean. Verified end to end on a Testnet machine: replacing the TXT record with the correctly derived token made the scheduler log `domain verification successful`, and the certificate was issued.


matevz:
This seems to be a regression I introduced in https://github.com/oasisprotocol/cli/pull/696.

build/rofl/scheduler/domain.go
```diff
@@ -28,9 +28,13 @@ const (
...
 // The result is then encoded using base64.
-func DomainVerificationToken(instance *roflmarket.Instance, appID rofl.AppID, domain string) string {
+//
+// Note that deployedAppID must be the app deployed on the instance
+// (instance.Deployment.AppID) and not the provider's scheduler app, otherwise the
+// scheduler will reject the resulting token.
+func DomainVerificationToken(instance *roflmarket.Instance, deployedAppID rofl.AppID, domain string) string {
 	h := tuplehash.New256(32, []byte(DomainVerificationTokenContext))
-	_, _ = h.Write(appID[:])
+	_, _ = h.Write(deployedAppID[:])
```
matevz: I think we can skip the new comment and `appID` renaming. The scheduler's app ID confusion was not deliberate in any way.

cmd/rofl/machine/show.go
```diff
@@ -223,12 +223,20 @@ func prettyPrintMachinePorts(extraCfg *roflCmdBuild.AppExtraConfig, appID rofl.A
...
 
-			domainToken := scheduler.DomainVerificationToken(insDsc, appID, p.CustomDomain)
 			addrs, err := net.LookupHost(genericDomain)
 			if err == nil {
 				fmt.Printf("      * Point A record of your domain to: %s\n", addrs[0])
 			}
-			fmt.Printf("      * Add TXT record to your domain: oasis-rofl-verification=%s\n", domainToken)
+
+			// The token is bound to the deployed app, so there is nothing to show
+			// before something is actually deployed on the machine.
```
matevz: Obvious from the printed notice below IMO.

```suggestion
```

matevz (approved):
Good job. I also tested it and seems to work.

build/rofl/scheduler/domain.go
anzoman: Reverted, `domain.go` is untouched now.

cmd/rofl/machine/show.go
anzoman: Dropped it.

